PR #9, opened 2022-11-23
[ci] Fix check-symbols failing on ubuntu-latest

swift-kim:
The ubuntu-latest image now maps to ubuntu-22.04 on GitHub Actions. Ignore the symbol version part of the `nm` output.

https://github.com/flutter-tizen/embedder/actions/runs/3528676880

### DIFF
--- a/tools/check_symbols.py
+++ b/tools/check_symbols.py
@@ -21,6 +21,7 @@ class Symbol:
 
   @staticmethod
   def parse(line):
+    # Format: "         U abort@GLIBC_2.4" (addr can be empty)
     return Symbol(line[:8].strip(), line[9], line[11:].strip().split('@')[0])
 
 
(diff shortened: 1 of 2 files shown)


swift-kim:
This change will take effect after being merged to the master branch. Please ignore the PR check.

JSUYA (approved):

tools/check_symbols.py
```diff
@@ -21,7 +21,7 @@ def __str__(self):
 
   @staticmethod
   def parse(line):
-    return Symbol(line[:8].strip(), line[9], line[11:].strip())
+    return Symbol(line[:8].strip(), line[9], line[11:].strip().split('@')[0])
```
WonyoungChoi, line 24: Could you add a comment line with an example of the line to parse?

WonyoungChoi (approved):

tools/check_symbols.py
swift-kim: Actually there's no real good example but I'll try.

PR #10, opened 2022-11-25
TizenRendererEgl: Use AnyCast() to cast to tbm_surface_queue

JSUYA:
Previously, there was a problem using Dali::AnyCast() due to libcxx issues.
The embedder has been migrated not to use flutter's libcxx(use_flutter_cxx option)
Therefore, there is no need to directly access the implementation of container of Dali::Any.
So it changed to cast to tbm_surface_queue_h using Dali::AnyCast().


JSUYA:
https://github.com/flutter-tizen/flutter-tizen/issues/351#issuecomment-1325922782

swift-kim (approved):

PR #16, opened 2022-12-13
TizenViewBase: Remove TizenViewType enum and GetType()

JSUYA:
The embedder does not require a type for the class because -rtti option is enabled.
So, remove the type enum and the getter method.

### DIFF
--- a/flutter/shell/platform/tizen/channels/platform_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_channel.cc
@@ -168,8 +168,7 @@ void PlatformChannel::HandleMethodCall(
 }
 
 void PlatformChannel::SystemNavigatorPop() {
-  auto* view = dynamic_cast<TizenView*>(view_);
-  if (view) {
+  if (auto* view = dynamic_cast<TizenView*>(view_)) {
     view->SetFocus(false);
   } else {
     ui_app_exit();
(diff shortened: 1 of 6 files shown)


flutter/shell/platform/tizen/flutter_tizen_view.cc
```diff
@@ -94,13 +90,13 @@ void FlutterTizenView::CreateRenderSurface(
 
   if (engine_ && engine_->renderer()) {
     TizenGeometry geometry = tizen_view_->GetGeometry();
-    if (tizen_view_->GetType() == TizenViewType::kWindow) {
-      auto* window = reinterpret_cast<TizenWindow*>(tizen_view_.get());
+    if (dynamic_cast<TizenWindow*>(tizen_view_.get())) {
```
swift-kim, line 93: `typeid` should be better in this specific case.

https://en.wikipedia.org/wiki/Run-time_type_information

> The use of `typeid`, in a non-polymorphic context, is often preferred over `dynamic_cast<class_type>` in situations where just the class information is needed, because `typeid` is always a [constant-time](https://en.wikipedia.org/wiki/Constant_time) procedure

flutter/shell/platform/tizen/flutter_tizen_view.cc
```diff
@@ -49,9 +49,8 @@ namespace flutter {
 FlutterTizenView::FlutterTizenView(std::unique_ptr<TizenViewBase> tizen_view)
     : tizen_view_(std::move(tizen_view)) {
   tizen_view_->SetView(this);
-
```
swift-kim, line 52: Removing this blank line will make the code less readable.

swift-kim (commented):
Thank you for the work.

Actually the use of RTTI is [against the style guide](https://google.github.io/styleguide/cppguide.html#Run-Time_Type_Information__RTTI_) as we discussed before but I think the new code is still better than what it used to be (a hand-implemented RTTI-like workaround).

flutter/shell/platform/tizen/flutter_tizen_view.cc
JSUYA: done

flutter/shell/platform/tizen/flutter_tizen_view.cc
JSUYA: > typeid should be better in this specific case.

Oh... The type of `tizen_view_` is `std::unique_ptr<TizenViewBase>`.
I think, `typeid()` didn't make sense in this case because it returns `std::unique_ptr<TizenViewBase>`.
The reason `dynamic_cast` is used here is that it returns null if `tizen_view_` cannot be casted to `TizenWindow*`.

swift-kim: Hmm? Does `typeid(tizen_view_.get())` return the typeid of `TizenViewBase`?

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -158,10 +158,11 @@ void PlatformChannel::HandleMethodCall(
 }
 
 void PlatformChannel::SystemNavigatorPop() {
-  if (view_->GetType() == TizenViewType::kWindow) {
-    ui_app_exit();
+  auto* view = dynamic_cast<TizenView*>(view_);
+  if (view) {
```
swift-kim, line 162: Note that this can be even simplified to
```suggestion
  if (auto* view = dynamic_cast<TizenView*>(view_)) {
```
if you prefer. (I don't know if it's a good practice in terms of code readability.)

JSUYA: I updated code style.
Actually, I don't know if this is more good for readability. However, I think it might(?) be good because it makes it easier to know the scope of local variables.

flutter/shell/platform/tizen/flutter_tizen_view.cc
JSUYA: I was test this
```cpp
FT_LOG(Error) << "CJS typeid(renderer_.get()) : "
                << typeid(renderer_.get()).name();
FT_LOG(Error) << "CJS typeid(TizenRendererEvasGL).name() : "
                << typeid(TizenRendererEvasGL).name();

auto& casted_renderer = *renderer_.get();
FT_LOG(Error) << "CJS typeid(casted_renderer) : "
              << typeid(casted_renderer).name();
FT_LOG(Error)
    << "CJS (typeid(casted_renderer) == typeid(TizenRendererEvasGL)) : "
    << (typeid(casted_renderer) == typeid(TizenRendererEvasGL));
```
```kernel
[E] flutter_tizen_engine.cc: RunEngine(163) > CJS typeid(renderer_.get()) : PN7flutter13TizenRendererE
[E] flutter_tizen_engine.cc: RunEngine(165) > CJS typeid(TizenRendererEvasGL).name() : N7flutter19TizenRendererEvasGLE
[E] flutter_tizen_engine.cc: RunEngine(169) > CJS typeid(casted_renderer) : N7flutter19TizenRendererEvasGLE
[E] flutter_tizen_engine.cc: RunEngine(171) > CJS (typeid(casted_renderer) == typeid(TizenRendererEvasGL)) : 1
```

So we can use it like this.
```cpp
auto& casted_renderer = *renderer_.get();
if (IsHeaded() && (typeid(casted_renderer) == typeid(TizenRendererEvasGL))) {
```
but, I think it's better now than creating a new local variable.

+)
`typeid(*renderer_.get())` is make  `-Werror,-Wpotentially-evaluated-expression`

flutter/shell/platform/tizen/flutter_tizen.cc
```diff
@@ -280,35 +280,33 @@ void FlutterDesktopViewOnKeyEvent(FlutterDesktopViewRef view,
...
+
+  if (auto* tizen_view = dynamic_cast<flutter::TizenViewNui*>(
+          ViewFromHandle(view)->tizen_view())) {
+    if (ViewFromHandle(view)->engine()->renderer()->type() ==
+        FlutterDesktopRendererType::kEGL) {
+      tizen_view->OnKey(device_name, device_class, device_subclass, key, string,
+                        nullptr, modifiers, scan_code, timestamp, is_down);
+    }
   }
+
```
swift-kim, line 292: No blank lines should be necessary here.

swift-kim (approved):

flutter/shell/platform/tizen/flutter_tizen.cc
JSUYA: oh! thank you.

swift-kim: Oh, I mean both 283 and 292nd lines.

JSUYA: oh.. thanks again

PR #21, opened 2023-01-27
[engine] Sync Flutter 3.7 source code

swift-kim:
Copy common headers and source code (including `embedder.h` and cpp_client_wrapper) from the Flutter 3.7 engine source tree except accessibility-related code.

TODO:
- [x] Fix the black screen and flickering issue.
  - There will be a follow-up PR soon.
- [x] Implement a thread-safe messenger (`FlutterDesktopMessengerAddRef` and etc.)
  - Refer to https://github.com/flutter/engine/pull/36909/files.
  - We don't currently have a plan to implement this. Let's revisit when it becomes necessary.
- [ ] Sync the accessibility-related code (including `third_party/accessibility`).
- [ ] Replace deprecated `update_semantics_node_callback` with `update_semantics_callback`.
  - Refer to https://github.com/flutter/engine/pull/37404/files.
- [ ] Migrate plugins to use `TextureRegistrar::UnregisterTexture(texture_id, callback)` instead of [deprecated](https://github.com/flutter/engine/blob/707458b9292446276af2552568ebeff6364adea9/shell/platform/common/client_wrapper/include/flutter/texture_registrar.h#L104) `TextureRegistrar::UnregisterTexture(texture_id)`.

PR #23, opened 2023-02-03
[a11y] Sync Flutter 3.7 source code

swift-kim:
- Sync `flutter/third_party/accessibility` and rebase Tizen commits.
- Sync accessibility-related files in `flutter/shell/platform/common` which were not included in https://github.com/flutter-tizen/embedder/pull/21.
- Update the embedder implementation to resolve a build error.
  - Rename `AccessibilityBridgeDelegateTizen` to `AccessibilityBridgeTizen` according to https://github.com/flutter/engine/pull/36597/files.


bbrto21 (approved):

JSUYA (approved):
Any reason to revert `Accessibility number formatting improvements for Windows`?
This patch has been relanded here. (https://github.com/flutter/engine/pull/30301)
I think this patch would improve the reading of the double type.

swift-kim:
@JSUYA A short answer: The build will fail if you apply the patch.

A detailed answer would be: The patch requires a dependency on `double-conversion` and adding the dependency means we're adding the whole Dart SDK (`third_party/dart`) to the embedder's DEPS. That's too expensive.

PR #24, opened 2023-02-03
[a11y] Move to new update semantics API

swift-kim:
Align the Tizen embedder with the new `update_semantics_callback` embedder API.

For details, please see https://github.com/flutter/engine/pull/37404. I referred to Windows and macOS implementations.

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen_engine.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_engine.cc
@@ -206,8 +206,11 @@ bool FlutterTizenEngine::RunEngine() {
     args.custom_dart_entrypoint = project_->custom_dart_entrypoint().c_str();
   }
 #ifndef WEARABLE_PROFILE
-  args.update_semantics_node_callback = OnUpdateSemanticsNode;
-  args.update_semantics_custom_action_callback = OnUpdateSemanticsCustomActions;
+  args.update_semantics_callback = [](const FlutterSemanticsUpdate* update,
+                                      void* user_data) {
+    auto* engine = static_cast<FlutterTizenEngine*>(user_data);
+    engine->OnUpdateSemantics(update);
+  };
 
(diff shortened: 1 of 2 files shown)


loic-sharma (approved):
LGTM

bbrto21 (approved):

JSUYA (approved):

loic-sharma:
@swift-kim Hello, we just realized that `FlutterSemanticsUpdate` is flawed: array indexing on `FlutterSemanticsUpdate.nodes` and `FlutterSemanticsUpdate.custom_actions` will break when we add new members to `FlutterSemanticsNode` or `FlutterSemanticsCustomAction`. Instead, you'll need to use the `FlutterSemanticsNode.struct_size` and `FlutterSemanticsCustomAction.struct_size` to determine the next element's pointer.

This code is bad:
```C++
FlutterSemanticsUpdate update = ...;
FlutterSemanticsNode first = update.nodes[0];
FlutterSemanticsNode second = update.nodes[1]; // BAD! This could crash.
```

You'll want something like this instead:

```C++
// Note: untested
FlutterSemanticsUpdate update = ...;
FlutterSemanticsNode first = update.nodes[0];
FlutterSemanticsNode* second = static_cast<FlutterSemanticsNode*>(
  static_cast<char*>(update.nodes) + 1 * first.struct_size);
```

I'd suggest reverting this change and staying on the old semantics update callbacks for now. Apologies for the inconvenience.

swift-kim:
@loic-sharma Thank you for informing us. We always stay on the Flutter SDK stable channel and the `embedder.h` and engine implementation will remain unchanged on our side until the next stable engine release. So the current code should work at least until then. I'll keep track of any changes to `embedder.h` and `FlutterSemanticsNode` and make necessary changes to our embedder when I update the engine next time, so you don't have to worry. Thanks again for your great work.

Issue #6, opened 2021-10-22
Advanced text input options


swift-kim:
https://api.flutter.dev/flutter/services/TextInputConfiguration-class.html

Needs investigation:
- [ ] `autocorrect`: When disabled, auto-correction is disabled. → `ECORE_IMF_INPUT_PANEL_LAYOUT_PASSWORD`
- [ ] `autofillConfiguration`: Setup and provide the initial text values and hints. → `Ecore_IMF_Input_Hints`? (We may need to implement `TextInput.requestAutofill` and `TextInput.finishAutofillContext` first.)
- [ ] `enableSuggestions`: When disabled, input suggestions are disabled. → `Ecore_IMF_Input_Hints`?
- [ ] `inputAction`: Display a certain kind of action button. → `Ecore_IMF_Input_Panel_Return_Key_Type`
- [ ] `obscureText`: When enabled, input suggestions are disabled. → `Ecore_IMF_Input_Hints`?
- [x] `textCapitalization`: Select an uppercase or lowercase keyboard. The value can be either `words`, `sentences`, `characters`, or `none`. → `Ecore_IMF_Autocapital_Type`

Not applicable for Tizen:
- `actionLabel`
- `enableIMEPersonalizedLearning`

See also:
- https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/services/text_input.dart
- https://github.com/flutter/engine/blob/master/shell/platform/android/io/flutter/embedding/engine/systemchannels/TextInputChannel.java
- https://github.com/flutter/engine/blob/master/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java

PR #25, opened 2023-02-09
Advanced text input options

Swanseo0:
Support text input options
- `enableSuggestions`
- `inputAction`
- `autofillConfiguration`

To support `TextInput.requestAutofill`, the following features have been implemented.
- `tizen autofill`
- `nui`, `elementay` type's popup for autofill (ecore wl2 type doesn't yet provide)

Related to issue #6


Swanseo0:
In order to compile NUI after this patch, need to use `--system-cxx`.

tools/generate_sysroot.py
```diff
@@ -113,6 +113,14 @@
   'wayland-devel',
 ]
 
+# Only available for Tizen 5.5 and above.
+autofill_packages = [
```
swift-kim, line 117: The embedder is already built for Tizen 5.5+. No need to separate these packages. The only exceptions are Dali2 packages which are only available for Tizen 6.5+.

swift-kim (changes requested):

flutter/shell/platform/tizen/BUILD.gn
```diff
@@ -174,15 +179,26 @@ template("embedder") {
     defines += invoker.defines
     defines += [ "FLUTTER_ENGINE_NO_PROTOTYPES" ]
 
-    if (api_version == "6.5" && target_name != "flutter_tizen_wearable") {
+    if (api_version == "6.5") {
```
swift-kim, line 182: `api_version` is expected to be either 5.5 or 6.5. No other values are allowed at the moment.

Swanseo0: I think I misunderstood it. it's modified.

tools/generate_sysroot.py
Swanseo0: Thank you. Removed unnecessary package separation.

flutter/shell/platform/tizen/BUILD.gn
```diff
@@ -143,18 +143,23 @@ template("embedder") {
...
+        "capi-ui-autofill",
+        "capi-ui-autofill-common",
         "ecore_wl2",
         "tdm-client",
         "tizen-extension-client",
         "EGL",
         "GLESv2",
       ]
+
+      defines += [ "AUTOFILL_SUPPORT" ]
```
JSUYA, line 162: https://docs.tizen.org/application/native/guides/text-input/autofill/
According to the documentation, it seems that the autoffill library can also be added to a wearable profile.
Is there a reason to exclude wearable profiles?

swift-kim: The module was added in Tizen 5.5, but the wearable embedder should support Tizen 4.0 as well.

Swanseo0: [Check symbols #56](https://github.com/flutter-tizen/embedder/actions/runs/4131073926/jobs/7138378335)
When the first PR was opened, APIs related to autofill obtained the results of `Symbols not allowed (artifacts/libflutter_tizen_wearable.so)`. So excluded.

swift-kim: @Swanseo0 By the way you can also just use the `WERABLE_PROFILE` macro without having to define another one if you want.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,183 @@
...
+    FT_LOG(Error) << "autofill_fill_request error";
+  }
+  autofill_view_info_destroy(view_info);
+
+  response_items_.clear();
+}
+
+void TizenAutofill::RegisterAutofillItem(std::string view_id,
+                                         AutofillItem item) {
+  autofill_save_item_h si_h;
```
JSUYA, line 122: How about `autofill_save_item` or `save_item` instead `si_h`?

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,183 @@
...
+  autofill_save_item_set_autofill_hint(si_h, item.hint_);
+  autofill_save_item_set_id(si_h, item.id_.c_str());
+  autofill_save_item_set_label(si_h, item.label_.c_str());
+  autofill_save_item_set_sensitive_data(si_h, item.sensitive_data_);
+  autofill_save_item_set_value(si_h, item.value_.c_str());
+
+  char* app_id;
+  app_get_id(&app_id);
+
+  autofill_save_view_info_h svi_h;
```
JSUYA, line 134: Please change it to an appropriate name

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,183 @@
...
+void TizenAutofill::InitailizeAutofill() {
+  autofill_create(&autofill_);
+
+  int ret = autofill_connect(
+      autofill_,
+      [](autofill_h ah, autofill_connection_status_e status, void* user_data) {
+      },
+      NULL);
+  if (ret != AUTOFILL_ERROR_NONE) {
+    FT_LOG(Error) << __FUNCTION__ << "connect_autofill_daemon error";
```
JSUYA, line 31: Please remove `__FUNCTION__` and change it with an appropriate message. (`autofill_connect()`)

flutter/shell/platform/tizen/BUILD.gn
JSUYA: @Swanseo0  Thank you for check. I also agree to use an other macro than `AUTOFILL_SUPPORT`.

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: Here's the guideline.

https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

> Minimize the use of abbreviations that would likely be unknown to someone outside your project (especially acronyms and initialisms).

Also make sure to initialize handles to `nullptr` whenever possible.

flutter/shell/platform/tizen/tizen_autofill.h
```diff
@@ -0,0 +1,74 @@
...
+
+  void RegisterAutofillItem(std::string view_id, AutofillItem item);
+
+  void StoreResponseItem(std::unique_ptr<AutofillItem> item) {
+    response_items_.push_back(move(item));
+  }
+
+  void SetPopupCallback(std::function<void()> callback) {
+    popup_callback_ = callback;
+  }
```
bbrto21, line 41: I think that using prefix 'On'  and remove postfix 'Callback' seems more consistent.
```c++
  void SetOnPopup(std::function<void()> on_popup) {
    on_popup_ = on_popup;
  }
```
...
```c++
  void SetOnCommit(std::function<void(std::string)> on_commit) {
      on_commit_ = on_commit;
  }
 
  void OnCommit(std::string str) { on_commit_(str); }
```

flutter/shell/platform/tizen/tizen_window_elementary.cc
```diff
@@ -426,6 +431,35 @@ void TizenWindowElementary::PrepareInputMethod() {
       [this]() { view_delegate_->OnComposeEnd(); });
   input_method_context_->SetOnCommit(
       [this](std::string str) { view_delegate_->OnCommit(str); });
+
+#ifdef AUTOFILL_SUPPORT
+  ctxpopup_ = elm_ctxpopup_add(elm_win_);
```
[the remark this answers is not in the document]
bbrto21: How about moving this line to `CreateView`?
also the 456 line callback registration part would look better if moved to `RegisterEventHandlers`.
If you agree, I would appreciate it if you could apply it to similar codes in other files.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,183 @@
...
+}
+
+void TizenAutofill::InitailizeAutofill() {
+  autofill_create(&autofill_);
+
+  int ret = autofill_connect(
+      autofill_,
+      [](autofill_h ah, autofill_connection_status_e status, void* user_data) {
+      },
+      NULL);
```
bbrto21, line 29:
```suggestion
      nullptr);
```

flutter/shell/platform/tizen/tizen_autofill.cc
bbrto21:
```suggestion
  autofill_save_view_info_h svi_h = nullptr;
```

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  autofill_save_item_create(&si_h);
+  autofill_save_item_set_autofill_hint(si_h, item.hint_);
+  autofill_save_item_set_id(si_h, item.id_.c_str());
+  autofill_save_item_set_label(si_h, item.label_.c_str());
+  autofill_save_item_set_sensitive_data(si_h, item.sensitive_data_);
+  autofill_save_item_set_value(si_h, item.value_.c_str());
+
+  char* app_id;
+  app_get_id(&app_id);
```
bbrto21, line 132: Since you do not check the return value of `app_get_id`, it is safe to initialize `app_id` as nullptr.
Also, the if statement in line 142 may also behave incorrectly when calling `app_get_id` failed.

flutter/shell/platform/tizen/tizen_window_elementary.cc
```diff
@@ -426,6 +431,35 @@ void TizenWindowElementary::PrepareInputMethod() {
...
+
+  input_method_context_->SetOnPopupAutofillContext([this]() {
+    if (TizenAutofill::GetInstance().GetAutofillItems().size() > 0) {
```
bbrto21, line 439: If clear all items before starting to add items, the `EVAS_CALLBACK_HIDE` for `ctxpopup_` no longer needed??

```suggestion
  input_method_context_->SetOnPopupAutofillContext([this]() {
    elm_ctxpopup_clear(ctxpopup_);
    if (TizenAutofill::GetInstance().GetAutofillItems().size() > 0) {
```

Swanseo0: If change it as you said, we won't need `EVAS_CALLBACK_HIDE`, but the previous information will remain until `requestAutofill` is executed.

Would it be better to implement it so that it doesn't use `EVAS_CALLBACK_HIDE`?

flutter/shell/platform/tizen/tizen_autofill.h
Swanseo0: Thank you for the good suggestion. it's modified.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: The variable name is modified to `save_view_info` and initialized with `nullptr`.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,82 @@
...
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::AUTO_HIDE_DELAY, 2500);
+}
+
+void NuiAutofillPopup::PopupAutofill(Dali::Actor* actor) {
+  const auto& items = TizenAutofill::GetInstance().GetAutofillItems();
+  if (items.size() > 0) {
+    PrepareAutofill();
+    Dali::Toolkit::TableView content =
+        Dali::Toolkit::TableView::New(items.size(), 1);
+    // content.SetCellPadding(Dali::Size(10.0f, 0));
```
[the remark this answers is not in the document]
Swanseo0: Deleted the line. thank you.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: `app_id` was initialized with `nullptr`.

flutter/shell/platform/tizen/tizen_window_elementary.cc
bbrto21: Um... Is there a path for the popup to be hidden other than when an item is selected? 
If so, I think it is good to clear the item through the `EVAS_CALLBACK_HIDE` callback.
also, when an item is selected, it would be nice to clear all items immediately after committing.

flutter/shell/platform/tizen/tizen_view_elementary.h
```diff
@@ -51,6 +51,10 @@ class TizenViewElementary : public TizenView {
   Evas_Object* parent_ = nullptr;
   Evas_Object* container_ = nullptr;
   Evas_Object* image_ = nullptr;
+#ifndef WEARABLE_PROFILE
+  Evas_Object* ctxpopup_ = nullptr;
+  Evas_Object_Event_Cb popup_hide_callback_ = nullptr;
```
bbrto21, line 56: How about using `evas_object_callbacks_` to store function pointer instead of this?

bbrto21 (commented):
Could you kindly guide me on how to test this feature? 
It would be even better if you could share the test app you used for development.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,81 @@
...
+  const Dali::PointState::Type state = event.GetState(0);
+  if (Dali::PointState::DOWN == state) {
+    auto t = actor.GetProperty(Dali::Actor::Property::NAME).Get<std::string>();
+    on_commit_(t);
+    OnHide();
+  }
+  return true;
+}
+
+void NuiAutofillPopup::OnHide() {
```
bbrto21, line 25: I Think this seems to make more sense. 

```suggestion
void NuiAutofillPopup::OutsideTouched() {
```

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,81 @@
...
+#include <dali-toolkit/public-api/controls/text-controls/text-label.h>
+
+#include "flutter/shell/platform/tizen/tizen_autofill.h"
+
+namespace flutter {
+bool NuiAutofillPopup::OnTouch(Dali::Actor actor,
+                               const Dali::TouchEvent& event) {
+  const Dali::PointState::Type state = event.GetState(0);
+  if (Dali::PointState::DOWN == state) {
+    auto t = actor.GetProperty(Dali::Actor::Property::NAME).Get<std::string>();
```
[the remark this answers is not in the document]
bbrto21: @Swanseo0 We're avoiding using auto as much as possible for readability :)

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,81 @@
...
+    on_commit_(t);
+    OnHide();
+  }
+  return true;
+}
+
+void NuiAutofillPopup::OnHide() {
+  // TODO : There is a phenomenon where white traces remain for a while when
```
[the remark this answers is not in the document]
Swanseo0: Added it. thank you.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,81 @@
...
+#include "flutter/shell/platform/tizen/tizen_autofill.h"
+
+namespace flutter {
+bool NuiAutofillPopup::OnTouch(Dali::Actor actor,
+                               const Dali::TouchEvent& event) {
+  const Dali::PointState::Type state = event.GetState(0);
+  if (Dali::PointState::DOWN == state) {
+    auto t = actor.GetProperty(Dali::Actor::Property::NAME).Get<std::string>();
+    on_commit_(t);
+    OnHide();
```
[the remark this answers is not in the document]
Swanseo0: Changed from `OnHide` to `Hide`.

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
Swanseo0: > What's t and what's its type? Could you be more specific?

Change from `t` to `text`. `text`'s type is `std::string`.

> @Swanseo0 We're avoiding using auto as much as possible for readability :)

For readability, i will avoid the use of auto too.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,81 @@
...
+
+#include "flutter/shell/platform/tizen/nui_autofill_popup.h"
+
+#include <dali-toolkit/dali-toolkit.h>
+#include <dali-toolkit/devel-api/controls/table-view/table-view.h>
+#include <dali-toolkit/public-api/controls/text-controls/text-label.h>
+
+#include "flutter/shell/platform/tizen/tizen_autofill.h"
+
+namespace flutter {
```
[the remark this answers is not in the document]
Swanseo0: Added a blank line. thank you.

flutter/shell/platform/tizen/channels/text_input_channel.h
```diff
@@ -79,6 +80,12 @@ class TextInputChannel {
   // Automatic text capitalization type. See available options:
   // https://api.flutter.dev/flutter/services/TextCapitalization.html
   std::string text_capitalization_ = "";
+
+  std::string autofill_id_;
+
+  // Hints for request autofill. See availabel options:
```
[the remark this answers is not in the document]
Swanseo0: Modified the comment on `autofill_hints_`.

flutter/shell/platform/tizen/channels/text_input_channel.h
```diff
@@ -79,6 +80,12 @@ class TextInputChannel {
   // Automatic text capitalization type. See available options:
   // https://api.flutter.dev/flutter/services/TextCapitalization.html
   std::string text_capitalization_ = "";
+
+  std::string autofill_id_;
```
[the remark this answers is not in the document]
Swanseo0: Added the comment on `autofill_id_`.

flutter/shell/platform/tizen/channels/text_input_channel.cc
```diff
@@ -273,6 +319,13 @@ void TextInputChannel::HandleMethodCall(
           cursor_offset);
     }
     SendStateUpdate();
+  } else if (method.compare(kRequestAutofillMethod) == 0) {
+#ifndef WEARABLE_PROFILE
+    TizenAutofill& instance = TizenAutofill::GetInstance();
+    instance.RequestAutofill(autofill_hints_, autofill_id_);
+#else
+    result->NotImplemented();
+#endif
```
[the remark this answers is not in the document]
Swanseo0: Added `#ifndef WEARABLE_PROFILE` condition to declaration of `kRequestAutofillMethod` and changed it to code that  suggested.

flutter/shell/platform/tizen/channels/text_input_channel.cc
```diff
@@ -211,6 +236,27 @@ void TextInputChannel::HandleMethodCall(
...
+          unique_identifier_iter->value.IsString()) {
+        autofill_id_ = unique_identifier_iter->value.GetString();
+      }
+
+      auto hints_iter = autofill_iter->value.FindMember(kHints);
+      if (hints_iter != autofill_iter->value.MemberEnd() &&
+          hints_iter->value.IsArray()) {
+        for (auto hint = hints_iter->value.GetArray().Begin();
+             hint != hints_iter->value.GetArray().End(); hint++) {
+          autofill_hints_.clear();
```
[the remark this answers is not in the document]
Swanseo0: Changed the location that cleans the vector. thank you.

flutter/shell/platform/tizen/channels/text_input_channel.cc
```diff
@@ -60,6 +69,12 @@ TextInputChannel::TextInputChannel(
              std::unique_ptr<MethodResult<rapidjson::Document>> result) {
         HandleMethodCall(call, std::move(result));
       });
+#ifndef WEARABLE_PROFILE
+  TizenAutofill& instance = TizenAutofill::GetInstance();
```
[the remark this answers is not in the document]
Swanseo0: Changed it to `autofill`.

flutter/shell/platform/tizen/channels/text_input_channel.cc
```diff
@@ -60,6 +69,12 @@ TextInputChannel::TextInputChannel(
              std::unique_ptr<MethodResult<rapidjson::Document>> result) {
         HandleMethodCall(call, std::move(result));
       });
+#ifndef WEARABLE_PROFILE
```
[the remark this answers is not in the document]
Swanseo0: Added a blank line between `SetMethodCallHandler` and `#ifndef`.

flutter/shell/platform/tizen/channels/text_input_channel.cc
```diff
@@ -8,6 +8,10 @@
 #include "flutter/shell/platform/tizen/flutter_tizen_engine.h"
 #include "flutter/shell/platform/tizen/logger.h"
 
+#ifndef WEARABLE_PROFILE
```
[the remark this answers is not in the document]
Swanseo0: Have deleted a blank line that separate includes.

flutter/shell/platform/tizen/nui_autofill_popup.cc
Swanseo0: I think it would be better to distinguish between `OutsideTouched` and `Hide`. added `OutsideTouched` function that call `Hide`.

flutter/shell/platform/tizen/tizen_view_elementary.h
Swanseo0: Thank you for the good offer. changed it to store in `evas_object_callbacks_`.

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
swift-kim: Thanks, but no space is needed after the parenthesis.

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
swift-kim: `OutsideTouched` which is a duplicate of `Hide` looks redundant. You can just use `Hide`.

flutter/shell/platform/tizen/tizen_window_elementary.cc
Swanseo0: When `ctxpopup` loses focus, `ctxpopup` is automatically hidden.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,90 @@
...
+  popup_.SetProperty(Dali::Actor::Property::NAME, "popup");
+  popup_.SetProperty(Dali::Actor::Property::PARENT_ORIGIN,
+                     Dali::ParentOrigin::TOP_LEFT);
+  popup_.SetProperty(Dali::Actor::Property::ANCHOR_POINT,
+                     Dali::AnchorPoint::TOP_LEFT);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::TAIL_VISIBILITY, false);
+  popup_.SetBackgroundColor(Dali::Color::WHITE_SMOKE);
+  popup_.OutsideTouchedSignal().Connect(this,
+                                        &NuiAutofillPopup::OutsideTouched);
+  popup_.HiddenSignal().Connect(this, &NuiAutofillPopup::Hidden);
```
swift-kim, line 53: Hmm. `Hide` and `Hidden` sound too similar. If I understand correctly, if `Hide` is called, the popup is closed (hid), and in turn the `Hidden` callback is called. Is this right?

One option is to remove the `Hide` method and inline the implementation within `Touched`.

Swanseo0: > Hmm. `Hide` and `Hidden` sound too similar. If I understand correctly, if `Hide` is called, the popup is closed (hid), and in turn the `Hidden` callback is called. Is this right?

Correct.

> One option is to remove the Hide method and inline the implementation within Touched.

I think it's a good option. I will modify it.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,90 @@
...
+                        Dali::Vector4(10, 10, 0, 0));
+    for (uint32_t i = 0; i < items.size(); ++i) {
+      Dali::Toolkit::TextLabel label =
+          Dali::Toolkit::TextLabel::New(items[i]->label_);
+      label.SetProperty(Dali::Actor::Property::NAME, items[i]->value_);
+      label.SetResizePolicy(Dali::ResizePolicy::DIMENSION_DEPENDENCY,
+                            Dali::Dimension::HEIGHT);
+      label.SetProperty(Dali::Toolkit::TextLabel::Property::TEXT_COLOR,
+                        Dali::Color::WHITE_SMOKE);
+      label.SetProperty(Dali::Toolkit::TextLabel::Property::POINT_SIZE, 7.0f);
```
swift-kim, line 77: Q. Are there any default values for these properties? It'd be nice if we could avoid using magic numbers.

flutter/shell/platform/tizen/nui_autofill_popup.h
```diff
@@ -0,0 +1,40 @@
...
+#define EMBEDDER_NUI_AUTOFILL_POPUP_H_
+
+#include <dali-toolkit/devel-api/controls/popup/popup.h>
+
+#include <functional>
+
+namespace flutter {
+
+using OnCommit = std::function<void(std::string str)>;
+using OnRendering = std::function<void()>;
```
swift-kim, line 15: Unused alias `OnRendering`.

flutter/shell/platform/tizen/nui_autofill_popup.h
```diff
@@ -0,0 +1,40 @@
...
+
+  void OutsideTouched();
+
+  void PrepareAutofill();
+
+  void PopupAutofill(Dali::Actor* actor);
+
+  void SetOnCommit(OnCommit callback) { on_commit_ = callback; }
+
+  bool Touched(Dali::Actor actor, const Dali::TouchEvent& event);
```
swift-kim, line 31: The callbacks can be made private.

flutter/shell/platform/tizen/nui_autofill_popup.h
```diff
@@ -0,0 +1,40 @@
...
+#include <dali-toolkit/devel-api/controls/popup/popup.h>
+
+#include <functional>
+
+namespace flutter {
+
+using OnCommit = std::function<void(std::string str)>;
+using OnRendering = std::function<void()>;
+
+class NuiAutofillPopup : public Dali::ConnectionTracker {
```
swift-kim, line 17: Why does this class inherit from `Dali::ConnectionTracker`?

flutter/shell/platform/tizen/nui_autofill_popup.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef EMBEDDER_NUI_AUTOFILL_POPUP_H_
+#define EMBEDDER_NUI_AUTOFILL_POPUP_H_
+
+#include <dali-toolkit/devel-api/controls/popup/popup.h>
+
+#include <functional>
+
+namespace flutter {
+
+using OnCommit = std::function<void(std::string str)>;
```
swift-kim, line 14:
```suggestion
using OnCommit = std::function<void(const std::string& str)>;
```

flutter/shell/platform/tizen/nui_autofill_popup.h
```diff
@@ -0,0 +1,40 @@
+// Copyright 2023 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef EMBEDDER_NUI_AUTOFILL_POPUP_H_
+#define EMBEDDER_NUI_AUTOFILL_POPUP_H_
+
+#include <dali-toolkit/devel-api/controls/popup/popup.h>
+
+#include <functional>
```
swift-kim, line 10: Missing includes?

```cpp
#include <dali/public-api/dali-core.h>

#include <string>
```

Please consult me or @bbrto21 on how to organize includes in the implementation file (`nui_autofill_popup.cc`) in a consistent way.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,188 @@
...
+  autofill_destroy(autofill_);
+}
+
+void TizenAutofill::InitailizeAutofill() {
+  autofill_create(&autofill_);
+
+  int ret = autofill_connect(
+      autofill_,
+      [](autofill_h autofill, autofill_connection_status_e status,
+         void* user_data) {},
```
swift-kim, line 28: Is this deliberately no-op? No need to check the current connection status?

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,188 @@
...
+      nullptr);
+  if (ret != AUTOFILL_ERROR_NONE) {
+    FT_LOG(Error) << "connect_autofill_daemon error";
```
swift-kim, line 31: All log messages should be complete sentences (which start with a capital letter and end with a period) and provide useful information. At least it should be like: "Failed to connect to the autofill daemon."

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  autofill_fill_response_set_received_cb(
```
swift-kim, line 34: No need to check the return value in this case?

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
Swanseo0: Have modified.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,90 @@
...
+                                        &NuiAutofillPopup::OutsideTouched);
+  popup_.HiddenSignal().Connect(this, &NuiAutofillPopup::Hidden);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::BACKING_ENABLED, false);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::AUTO_HIDE_DELAY, 2500);
+}
+
+void NuiAutofillPopup::PopupAutofill(Dali::Actor* actor) {
+  const std::vector<std::unique_ptr<AutofillItem>>& items =
+      TizenAutofill::GetInstance().GetAutofillItems();
+  if (items.size() > 0) {
```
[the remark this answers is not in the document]
Swanseo0: Thank you for good suggestion. Modified to use `empty()`.

flutter/shell/platform/tizen/nui_autofill_popup.cc
Swanseo0: The default `text color` is black and the `point size` is about 15.
The background color of the pop-up window is dark blue, so I can't see the letters well, so I set it to white smoke.
The font size is set to an appropriate size because the default value is large, but we will modify it so that we do not use the magic number.
However, if we do not change the background color of popup, white smoke will be better than the basic color(black).

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: Deleted it.

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: Moved `Hidden`, `OutsideTouched`, `Touched` to private region.

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: The reason for inheriting `Dali::ConnectionTracker` in `NuiAutofillPopup` is to use `Connect` function in `Dali::Signal`.

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: Modified.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Modified the log comment.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: `connected_` have been added to check connection status.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Changed it to check the return value of this case.
Additionally, i have added a `intialized_` variable to save whether the initialization process has all been successfully performed.

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,86 @@
...
+                     Dali::ParentOrigin::TOP_LEFT);
+  popup_.SetProperty(Dali::Actor::Property::ANCHOR_POINT,
+                     Dali::AnchorPoint::TOP_LEFT);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::TAIL_VISIBILITY, false);
+  popup_.SetBackgroundColor(Dali::Color::WHITE_SMOKE);
+  popup_.OutsideTouchedSignal().Connect(this,
+                                        &NuiAutofillPopup::OutsideTouched);
+  popup_.HiddenSignal().Connect(this, &NuiAutofillPopup::Hidden);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::BACKING_ENABLED, false);
+  popup_.SetProperty(Dali::Toolkit::Popup::Property::AUTO_HIDE_DELAY, 2500);
```
JSUYA, line 51: Is AUTO_HIDE_DELAY an intended property?
If a value is set, it will automatically hide after that amount of time.
I think, this value looks a property used for toast popup.
If this is the intended concept, please ignore this comment.

flutter/shell/platform/tizen/nui_autofill_popup.cc
JSUYA: Does "white background" and "7.0f font size" look good for both tv and common (rpi) profiles?

 +) NUI popup may differ from Dali popup. It might be better to set it equal to that value.
https://docs.tizen.org/application/dotnet/guides/user-interface/nui/nui-components/Popup/

flutter/shell/platform/tizen/nui_autofill_popup.cc
```diff
@@ -0,0 +1,86 @@
...
+                            Dali::Dimension::HEIGHT);
+      label.SetProperty(Dali::Toolkit::TextLabel::Property::TEXT_COLOR,
+                        Dali::Color::WHITE_SMOKE);
+      label.SetProperty(Dali::Toolkit::TextLabel::Property::POINT_SIZE, 7.0f);
+      label.TouchedSignal().Connect(this, &NuiAutofillPopup::Touched);
+      content.AddChild(label, Dali::Toolkit::TableView::CellPosition(i, 0));
+      content.SetFitHeight(i);
+    }
+    popup_.SetProperty(Dali::Actor::Property::SIZE,
+                       Dali::Vector2(140.0f, 35.0f * items.size()));
```
JSUYA, line 79: The more items, the longer the popup.
I think it should be scrollable after a certain number, what do you think?

flutter/shell/platform/tizen/tizen_view_elementary.h
```diff
@@ -51,6 +51,9 @@ class TizenViewElementary : public TizenView {
   Evas_Object* parent_ = nullptr;
   Evas_Object* container_ = nullptr;
   Evas_Object* image_ = nullptr;
+#ifndef WEARABLE_PROFILE
+  Evas_Object* ctxpopup_ = nullptr;
```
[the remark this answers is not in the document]
Swanseo0: It looks better to change `autofill_popup_` than `ctxpopup_`. Thank you for a good idea.

flutter/shell/platform/tizen/nui_autofill_popup.cc
Swanseo0: This Property is intended.
At first, the situation when pop-up was hidden was as follows.
- Select an item
- Press Back
- Touch outside

However, when `popup_.SetProperty(Dali::Toolkit::Popup::Property::BACKING_ENABLED, false)` was added, `OutsideTouched` was not executed. So I added `AUTO_HIDE_DELAY`.

flutter/shell/platform/tizen/nui_autofill_popup.cc
Swanseo0: > Does "white background" and "7.0f font size" look good for both tv and common (rpi) profiles?

Reducing the size was better than the default.

> +) NUI popup may differ from Dali popup. It might be better to set it equal to that value.

In order to make Dali's popup the same as NUI's popup, I think we need to find a way to set Property in more detail. I tried to make it the same shape, but I couldn't find a way to set it up, such as the shape of popup, the border of each item, and etc.

> I think it should be scrollable after a certain number, what do you think?

I agree. I tried to make it scrollable, but... I couldn't find a way.

Swanseo0:
We can test using this [autofill sample](https://github.com/Swanseo0/autofill).
Yu can follow the steps below for the test method.
 - autofill_service app install
 - Settings - Language and input(System) - Autofill Service(INPUT ASSISTANCE) - autofill_service 선택
 - Running the app
 - text input field click

Four items with AUTOFILL_HINT_ID have been added from `runner.cc` in `autofill_service`.
Please refer to the creator of `TizenAutofillService`.

flutter/shell/platform/tizen/nui_autofill_popup.h
swift-kim: `Prepare` could also be private.

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
swift-kim: Hmm. I think `empty()` with `return` should be more readable than `!empty()`.

flutter/shell/platform/tizen/nui_autofill_popup.h
swift-kim: Wait, did you delete `OnCommit`?

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: `is_connected_` is usually better than just `connected_`.

https://dart.dev/guides/language/effective-dart/design#prefer-a-non-imperative-verb-phrase-for-a-boolean-property-or-variable

swift-kim: You shouldn't need the extra `SetConnected` method. You can directly access private member fields via `user_data`.

Also you shouldn't assume the `status` is always `AUTOFILL_CONNECTION_STATUS_CONNECTED`.

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: "Failed to" but not "Fail to".

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: You should not immediately `return` after `Initailize`.

Also `Initailize`→`Initialize`.

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: The `if (!connected_)` check will fail if it comes immediately after `Initailize()` and `autofill_connect` is not a synchronous call.

Related to the comment https://github.com/flutter-tizen/embedder/pull/25#discussion_r1105334997.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,216 @@
...
+
+  int ret = autofill_commit(autofill_, save_view_info);
+  if (ret != AUTOFILL_ERROR_NONE) {
+    FT_LOG(Error) << "Fail to register autofill item.";
+  }
+
+  autofill_save_view_info_destroy(save_view_info);
+}
+
+std::optional<autofill_hint_e> TizenAutofill::ConvertAutofillHint(
```
swift-kim, line 187: This method doesn't have to be a class member. Instead, you can create an anonymous namespace at the top of this file and move this function into it. (It's effectively the same as creating a C-style `static` function.)

flutter/shell/platform/tizen/tizen_autofill.h
```diff
@@ -0,0 +1,77 @@
...
+  autofill_h autofill_ = nullptr;
+
+  std::vector<std::unique_ptr<AutofillItem>> response_items_;
+
+  std::function<void()> on_popup_;
+
+  std::function<void(const std::string&)> on_commit_;
+};
+
+#endif
```
swift-kim, line 77:
```suggestion
#endif  // EMBEDDER_TIZEN_AUTOFILL_H_
```

Also you need to enclose the entire class (lines 16-75) with the `flutter` namespace.

Swanseo0: Added comments and `flutter` namespace.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Created an anonoymous namespace at the top of `tizen_autofill.cc` and moved `ConvertAutofillHint` into it.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,216 @@
...
+    FT_LOG(Error) << "Fail to set fill response received callback.";
+    return;
+  }
+
+  response_items_.clear();
+  initailzed_ = true;
+}
+
+void TizenAutofill::RequestAutofill(std::vector<std::string> hints,
+                                    std::string id) {
```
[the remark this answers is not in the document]
Swanseo0: Added. Thank you!

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,216 @@
...
+                    if (label) {
+                      free(label);
+                    }
+
+                    return true;
+                  },
+                  group);
+              return true;
+            },
+            &count);
```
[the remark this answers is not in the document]
Swanseo0: By passing `this` as `user_data`, modified to not use `TizenAutofill::GetInstance()`.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  response_items_.clear();
+  initailzed_ = true;
+}
+
+void TizenAutofill::RequestAutofill(std::vector<std::string> hints,
+                                    std::string id) {
+  if (!initailzed_) {
+    Initailize();
```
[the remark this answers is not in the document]
Swanseo0: `Initialize` failed case is when `autofill_create` or `autofill_connet` failed.
In this case, `TizenAutofill` is not expected to work at all, so I think it needs to be reinitialized.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: > You should not immediately `return` after `Initailize`.

The reason for immediate `return` after `Initialize` is that `connect` is asynchronous.
If proceed with the autofill request immediately after `initialize`, it will fail because it is not connected.

> Also Initailize→Initialize.

Modified.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,216 @@
...
+    return;
+  }
+
+  ret = autofill_fill_response_set_received_cb(
+      autofill_,
+      [](autofill_h autofill, autofill_fill_response_h fill_response,
+         void* data) {
+        int count = 0;
+        autofill_fill_response_get_group_count(fill_response, &count);
+        autofill_fill_response_foreach_group(
```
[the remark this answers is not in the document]
Swanseo0: > You could make each nested callback into a private static method (or a named lambda expression) to reduce the level of indentation.

Reduced the level of identation.

> Otherwise, you could wrap some of the C functions into individual methods (in the sense of OOP) but I'm not sure if it's appropriate. You may redesign this entire class if you feel it's appropriate.

C functions with specific purposes were made into static functions.

> Also is it okay to just ignore the return values (`ret`) of these functions?

Added conditions for functions that need to be checked for `ret`.
I think other functions doesn't need to be checked, but is there anything I missed?

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: Moved `Prepare` to private region. thank you.

flutter/shell/platform/tizen/nui_autofill_popup.h
Swanseo0: I deleted it because there was a conflict with this while adding `const &`.
https://github.com/flutter-tizen/embedder/blob/a152671a2226bec8e84d70a06c0aa05226eb8bc7/flutter/shell/platform/tizen/tizen_input_method_context.h#L18

Should I add `OnCommit` again and apply `const &` to the above either?

flutter/shell/platform/tizen/nui_autofill_popup.cc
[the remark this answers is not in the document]
Swanseo0: Changed to `return` if `empty()`.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: > `is_connected_` is usually better than just `connected_`.

Modified to use `is_connected_`.
> You shouldn't need the extra `SetConnected` method. You can directly access private member fields via `user_data`.

Can I access the private region by casting `user_data` to `TizenAutofill`?
> Also you shouldn't assume the `status` is always `AUTOFILL_CONNECTION_STATUS_CONNECTED`.

Changed to call `SetConnected` according to `state`.
> The if (!connected_) check will fail if it comes immediately after Initailize() and autofill_connect is not a synchronous call.

`Autofill` will not work until connected. So we need to make sure it's connected.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Modified to "Failed to".

flutter/shell/platform/tizen/nui_autofill_popup.h
swift-kim: Yes, using an alias should be more readable. You may update `OnPreeditChanged` as well.

cc @bbrto21

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: > Can I access the private region by casting `user_data` to `TizenAutofill`?

Yes, you can. Also please do not rename `user_data` to `data` unless it is appropriate. We often use `self` as the name of the casted pointer.

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: That's the point of my comment above. The `RequestAutofill` method has indeterminate behavior depending on whether the `TizenAutofill` is initialized or not. If it's not initialized (although it's unlikely in normal situations), the method will simply call `Initialize` and exit. If it is already initialized, the rest of the method body will be executed.

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
swift-kim: Isn't `Initialize` likely to fail again if it failed before?

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
swift-kim: You don't have to apply this to pointer types like `autofill_view_info_h`.

Also you missed an `&` in line 136.

flutter/shell/platform/tizen/tizen_autofill.h
swift-kim: The same applies to `nui_autofill_popup.h`.

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
swift-kim: > Reduced the level of identation.

You may prefer static member methods over static functions, to allow accessing non-public members within callbacks.

> I think other functions doesn't need to be checked,

Is it because you think you can simply "ignore" if they fail? Well, that's quite uncommon.

flutter/shell/platform/tizen/tizen_autofill.h
```diff
@@ -0,0 +1,80 @@
...
+
+namespace flutter {
+
+struct AutofillItem {
+  autofill_hint_e hint_;
+  bool sensitive_data_;
+  std::string label_;
+  std::string id_;
+  std::string value_;
+};
```
swift-kim, line 24: Struct members should not be named like class members.

https://google.github.io/styleguide/cppguide.html#Variable_Names

Also, are the values other than `value_` ever used?

flutter/shell/platform/tizen/tizen_view_elementary.cc
```diff
@@ -354,6 +375,25 @@ void TizenViewElementary::PrepareInputMethod() {
...
+            [](void* data, Evas_Object* obj, void* event_info) {
+              AutofillItem* item = static_cast<AutofillItem*>(data);
+              TizenAutofill::GetInstance().OnCommit(item->value_);
+              evas_object_hide(obj);
+            },
+            item.get());
+      }
+    }
+    // TODO(Swanseo0): Change ctxpopup's position to focused input field.
+    evas_object_move(ctxpopup_, 0, 0);
```
swift-kim, line 393: Isn't the popup supposed to appear just above the keyboard?

tools/generate_sysroot.py
```diff
@@ -155,6 +159,7 @@ def generate_sysroot(sysroot: Path, api_version: float, arch: str, quiet=False):
   existing_rpms = [f for f in download_path.iterdir() if f.suffix == '.rpm']
 
   packages = base_packages + unified_packages
+
```
swift-kim, line 162: nit: Unexpected newline.

bbrto21:
> We can test using this [autofill sample](https://github.com/Swanseo0/autofill).

This link is broken.

tools/generate_sysroot.py
Swanseo0: Removed blank line.

flutter/shell/platform/tizen/tizen_view_elementary.cc
Swanseo0: We can set it to appear just above the input panel. Should I temporarily set it to appear just above the input panel?

flutter/shell/platform/tizen/tizen_autofill.h
Swanseo0: Mainly used are `hint`, `label`, and `value`.
`sensitive_data` and `id` are variable that are not used in `RequestAutoflll`. I think it will only be used in `RegisterItem`.

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,271 @@
...
+  if (ret != AUTOFILL_ERROR_NONE) {
+    FT_LOG(Error) << "Failed to request autofill.";
+  }
+  autofill_view_info_destroy(view_info);
+
+  response_items_.clear();
+}
+
+void TizenAutofill::RegisterItem(const std::string& view_id,
+                                 const AutofillItem& item) {
```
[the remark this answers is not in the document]
Swanseo0: It is not currently in use, but it is a function that was implemented because I thought it would be needed someday. Was it an unnecessary implementation?

flutter/shell/platform/tizen/tizen_autofill.cc
```diff
@@ -0,0 +1,271 @@
...
+  autofill_fill_response_item_get_presentation_text(item, &label);
+  autofill_fill_response_item_get_value(item, &value);
+  std::unique_ptr<AutofillItem> response_item =
+      std::make_unique<AutofillItem>();
+  response_item->id_ = std::string(id);
+  response_item->value_ = std::string(value);
+  response_item->label_ = std::string(label);
+
+  TizenAutofill* tizen_autofill = static_cast<TizenAutofill*>(data);
+  tizen_autofill->StoreResponseItem(move(response_item));
```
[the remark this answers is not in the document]
Swanseo0: Modified.

flutter/shell/platform/tizen/tizen_view_elementary.cc
```diff
@@ -94,6 +97,9 @@ bool TizenViewElementary::CreateView() {
   evas_object_image_size_set(image_, initial_width_, initial_height_);
   evas_object_image_alpha_set(image_, EINA_TRUE);
   elm_object_part_content_set(container_, "overlay", image_);
+#ifndef WEARABLE_PROFILE
+  ctxpopup_ = elm_ctxpopup_add(container_);
+#endif
```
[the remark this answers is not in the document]
Swanseo0: Added blank lines to separate each block.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Thank you for advice.
The `data` was changed to `user_data` again and use `self` as the name of the casted pointer.

flutter/shell/platform/tizen/tizen_autofill.cc
Swanseo0: Sorry for the late reply.
The reason you said "You should not immediate return after initiation" is because the connection is asynchronous, and we need to wait for the connection instead of `return` right away? Or should we not call the `Initialize` function at `RequestAutofill` function?

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
Swanseo0: When it fails and tries again, it seems likely to fail again. So I changed it so that it doesn't try `Initialize` again.

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
Swanseo0: The function that does not check the return value based on the current code is if the return value is `AUTOFILL_ERROR_NONE` or `AUTOFILL_ERROR_INVALID_PARAMETER`.
In the case of `AUTOFILL_ERROR_INVALID_PARAMETER`, I thought it could be ignored because the appropriate function was called with the appropriate parameters. Do I need to check this case?

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
Swanseo0: I modified it. Thank you

flutter/shell/platform/tizen/tizen_input_method_context.cc
```diff
@@ -273,6 +273,38 @@ bool TizenInputMethodContext::IsInputPanelShown() {
...
+  ecore_imf_context_prediction_allow_set(imf_context_,
+                                         enable ? EINA_TRUE : EINA_FALSE);
+}
+
+void TizenInputMethodContext::SetInputAction(const std::string& input_action) {
+  FT_ASSERT(imf_context_);
+  Ecore_IMF_Input_Panel_Return_Key_Type return_key_type =
+      ECORE_IMF_INPUT_PANEL_RETURN_KEY_TYPE_DEFAULT;
+
+  // Not support : none, previous, continueAction, route, emergencycall, newline
```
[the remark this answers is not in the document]
Swanseo0: Modified.

flutter/shell/platform/tizen/tizen_input_method_context.cc
```diff
@@ -273,6 +273,38 @@ bool TizenInputMethodContext::IsInputPanelShown() {
   return state == ECORE_IMF_INPUT_PANEL_STATE_SHOW;
 }
 
+void TizenInputMethodContext::SetEnableSuggestions(bool enable) {
+  FT_ASSERT(imf_context_);
+  ecore_imf_context_prediction_allow_set(imf_context_,
+                                         enable ? EINA_TRUE : EINA_FALSE);
```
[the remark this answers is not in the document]
Swanseo0: `SetContextOptions` was removed. and redundant function calls in `SetInputPanelOptions` function were removed and renamed to `SetInputPanelOption`.

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
Swanseo0: `RegisterItem` function was deleted.

flutter/shell/platform/tizen/tizen_autofill.cc
swift-kim: > The reason you said "You should not immediate return after initiation" is because the connection is asynchronous, and we need to wait for the connection instead of `return` right away?

Kind of yes.

> Or should we not call the `Initialize` function at `RequestAutofill` function?

I don't know. That's what you have to _think_ in the design phase.

flutter/shell/platform/tizen/tizen_autofill.cc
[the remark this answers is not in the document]
swift-kim: It depends. If you're sure the functions won't raise any errors or the errors can be ignored, error checking might not be necessary.

PR #30, opened 2023-04-14
Handle right and middle mouse button clicks

swift-kim:
Contributes to https://github.com/flutter-tizen/flutter-tizen/issues/490.

- Specify `FlutterPointerEvent.device_kind` according to the actual device type.
- Handle multi device states correctly.
- Handle mouse hover events and right and middle click events correctly.
- Fix `Evas_Event_Mouse_Down.button` being incorrectly used as a device ID and use `mouse_event->dev` instead.
- Update `kScrollOffsetMultiplier` with the value used by Chromium/Linux GTK.

The code added to `flutter_tizen_view.cc` is mostly based on the Windows implementation (`flutter_windows_view.cc`).

To test right and middle clicks, you can use the `GestureDetector` widget (`onSecondaryTapDown`/`onTertiaryTapDown`).

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen_view.h
+++ b/flutter/shell/platform/tizen/flutter_tizen_view.h
@@ -113,9 +113,6 @@ class FlutterTizenView : public TizenViewEventHandlerDelegate {
   // track of which buttons have been pressed, so it's the embedder's
   // responsibility.
   struct PointerState {
-    // The device kind.
-    FlutterPointerDeviceKind device_kind = kFlutterPointerDeviceKindMouse;
-
     // The device ID.
     int32_t device_id = 0;
 
--- a/flutter/shell/platform/tizen/flutter_tizen_view.h
+++ b/flutter/shell/platform/tizen/flutter_tizen_view.h
@@ -64,18 +64,21 @@ class FlutterTizenView : public TizenViewEventHandlerDelegate {
   void OnPointerMove(double x,
                      double y,
                      size_t timestamp,
+                     FlutterPointerDeviceKind device_kind,
                      int32_t device_id) override;
 
   void OnPointerDown(double x,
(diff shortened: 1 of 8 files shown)


flutter/shell/platform/tizen/flutter_tizen_view.h
```diff
@@ -112,6 +113,35 @@ class FlutterTizenView : public TizenViewEventHandlerDelegate {
...
+    // A virtual pointer ID that is unique across all device kinds.
+    int32_t pointer_id = 0;
+
+    // True if the last event sent to Flutter had at least one button pressed.
+    bool flutter_state_is_down = false;
+
+    // True if kAdd has been sent to Flutter. Used to determine whether
+    // to send a kAdd event before sending an incoming pointer event, since
+    // Flutter expects pointers to be added before events are sent for them.
+    bool flutter_state_is_added = false;
```
JSUYA, line 132: I think flutter_state_* is redundant. How about `is_down`, `is_added`, not flutter_state_*?

JSUYA (approved):

flutter/shell/platform/tizen/flutter_tizen_view.h
swift-kim: The design is from the [Windows embedder](https://github.com/flutter/engine/blob/main/shell/platform/windows/flutter_windows_view.h).

I'd keep using the `flutter_state_` prefix since it makes the meaning more explicit and clearer. Actually `is_down` has slightly different meanings for Flutter and the embedder. For Flutter, as you can see in the `FlutterPointerPhase` documentation, the down state means that _any_ (one or more) of the mouse buttons are pressed. In the embedder, however, we track the state of each mouse button individually, so if you just say `is_down`, I have to ask "which button is pressed?"

PR #31, opened 2023-04-17
TizenWindow,View: Add PointerDeviceKindMouse according to device class

JSUYA:
In Ecore_event and Evas_event, the devcie class can check Touch and Mouse event information.
So use this to pass the appropriate PointerDeviceKind.


JSUYA:
I have yet to find a way to check between Mouse and Touch events in NUI.

swift-kim:
Already implemented in https://github.com/flutter-tizen/embedder/pull/30.

PR #32, opened 2023-04-20
Implement the mouse cursor channel

JSUYA:
Change the ecore cursor theme and name
according to the kind of SystemMouseCursor received through the cursor channel.

Contributes to https://github.com/flutter-tizen/flutter-tizen/issues/490.


JSUYA:
This change only works with `TizenWindowEcoreWl2` + TV Profile. Haven't checked the behavior of `ecore_wl2_input_cursor_theme_name_set` in other environments yet.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,73 @@ void TizenWindowEcoreWl2::Show() {
   ecore_wl2_window_show(ecore_wl2_window_);
 }
 
+void TizenWindowEcoreWl2::UpdateFlutterCursor(const std::string& cursor_name) {
+#ifdef TV_PROFILE
+  int pointer_size = -1;
+  if (vconf_get_int(kSysMouseCursorPointerSizeVConfKey, &pointer_size) < 0) {
```
JSUYA, line 485: pointer_size will be 1.
If there is an option to change the mouse pointer size in options of the device, you can change this value.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,73 @@ void TizenWindowEcoreWl2::Show() {
...
+    FT_LOG(Error) << "Failed to load cursor size.";
```
swift-kim, line 486: The failure should not be logged as an error since there are devices that do not support the feature.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,73 @@ void TizenWindowEcoreWl2::Show() {
...
+  }
+
+  std::string input_cursor_name;
+  if (cursor_name == "basic") {
```
swift-kim, line 490: nit: `kind` is the argument name passed by the framework so this should fairly make sense:
```suggestion
  std::string cursor_name;
  if (kind == "basic") {
```

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,73 @@ void TizenWindowEcoreWl2::Show() {
...
+    switch (pointer_size) {
+      case 0:  // Large.
+        input_cursor_name = "large_normal";
+        break;
```
swift-kim, line 494: nit: How about not using the switch statement? You will have less number of lines with simple if-else and the code might be more readable.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
swift-kim: Can I configure this value in the system settings?

flutter/shell/platform/tizen/BUILD.gn
```diff
@@ -81,6 +82,7 @@ template("embedder") {
       "channels/accessibility_channel.cc",
       "channels/app_control.cc",
       "channels/app_control_channel.cc",
+      "channels/cursor_channel.cc",
```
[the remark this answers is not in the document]
JSUYA: The implementation of the cursor_channel class is taken from the window implementation.
In window, the cursor channel declared as cursor_handler.
We did not use _handler in channel class, so I used cursor_channel.

flutter/shell/platform/tizen/channels/cursor_channel.cc
```diff
@@ -0,0 +1,54 @@
...
+}  // namespace
+
+CursorChannel::CursorChannel(BinaryMessenger* messenger, TizenViewBase* view)
+    : view_(view) {
+  channel_ = std::make_unique<MethodChannel<EncodableValue>>(
+      messenger, kChannelName, &StandardMethodCodec::GetInstance());
+  channel_->SetMethodCallHandler(
+      [this](const MethodCall<EncodableValue>& call,
+             std::unique_ptr<MethodResult<EncodableValue>> result) {
+        this->HandleMethodCall(call, std::move(result));
```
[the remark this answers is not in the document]
JSUYA: done

flutter/shell/platform/tizen/channels/cursor_channel.h
```diff
@@ -0,0 +1,36 @@
...
+#define EMBEDDER_CURSOR_CHANNEL_H_
+
+#include "flutter/shell/platform/common/client_wrapper/include/flutter/binary_messenger.h"
+#include "flutter/shell/platform/common/client_wrapper/include/flutter/encodable_value.h"
+#include "flutter/shell/platform/common/client_wrapper/include/flutter/method_channel.h"
+#include "flutter/shell/platform/tizen/tizen_view_base.h"
+
+namespace flutter {
+
+// Channel to get/set the system's cursor type.
```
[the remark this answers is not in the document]
JSUYA: done

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,73 @@ void TizenWindowEcoreWl2::Show() {
...
+        input_cursor_name = "small_normal_input_field";
+        break;
+      default:
+        input_cursor_name = "normal_input_field";
+        break;
+    }
+  } else if (cursor_name == "none") {
+    input_cursor_name = "normal_transparent";
+  } else {
+    FT_LOG(Error) << cursor_name << " cursor is not supported.";
```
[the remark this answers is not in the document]
JSUYA: Added `return`. I think cursors not in the our list should not be set.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: Changed to `Info`

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: done

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: done

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: If there is no size option in System setting, you can edit vconf manually.
`vconftool set -f -t int db/menu/system/mouse-pointer-size 1`

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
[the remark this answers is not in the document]
JSUYA: Based on your review, I set cusor_name an initial value to be "normal_default".

flutter/shell/platform/tizen/BUILD.gn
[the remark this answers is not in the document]
swift-kim: But the channel name is the mouseCursor channel, and all other platform implementations (Android, darwin, Linux) use "mouse cursor" in their corresponding class names.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,58 @@ void TizenWindowEcoreWl2::Show() {
...
+    FT_LOG(Info) << "Failed to load cursor size.";
+  }
+
+  std::string cursor_name = "normal_default";
+  if (kind == "basic") {
+    if (pointer_size == 0) {  // Large.
+      cursor_name = "large_normal";
+    } else if (pointer_size == 1) {  // Medium.
+      cursor_name = "medium_normal";
+    } else if (pointer_size == 2) {  // Large.
```
swift-kim, line 495:
```suggestion
    } else if (pointer_size == 2) {  // Small.
```

flutter/shell/platform/tizen/tizen_view_elementary.cc
```diff
@@ -367,6 +367,10 @@ void TizenViewElementary::Show() {
   evas_object_show(image_);
 }
 
+void TizenViewElementary::UpdateFlutterCursor(const std::string& kind) {
+  FT_LOG(Error) << "UpdateFlutterCursor is not supported.";
```
swift-kim, line 371: This will result in too verbose error output on non-TV devices. I think the log level can be lowered to Info since the cursor type support is usually not that critical.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -469,6 +479,58 @@ void TizenWindowEcoreWl2::Show() {
...
+      cursor_name = "large_normal_input_field";
+    } else if (pointer_size == 1) {  // Medium.
+      cursor_name = "medium_normal_input_field";
+    } else if (pointer_size == 2) {  // Large.
+      cursor_name = "small_normal_input_field";
+    } else {
+      cursor_name = "normal_input_field";
+    }
+  } else if (kind == "none") {
+    cursor_name = "normal_transparent";
```
swift-kim, line 521: Is there any other way to unshow the cursor? It would be helpful if we could support hiding the cursor (`SystemMouseCursors.none`) on non-TV profiles so that users can use their own software cursors.

flutter/shell/platform/tizen/BUILD.gn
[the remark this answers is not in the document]
JSUYA: done

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: done

flutter/shell/platform/tizen/tizen_view_elementary.cc
JSUYA: Changed to `Info`.

swift-kim (approved):
Thank you!

PR #33, opened 2023-04-26
PlatformView: Add Offset method

Swanseo0:
Added Offset function to PlatformVIew to support [[webview_flutter_tizen] Fix youtube player on TV_PROFILE ](https://github.com/flutter-tizen/plugins/pull/564).

### DIFF
--- a/flutter/shell/platform/tizen/channels/platform_view_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_view_channel.cc
@@ -108,6 +108,8 @@ void PlatformViewChannel::HandleMethodCall(
     OnResize(arguments, std::move(result));
   } else if (method == "touch") {
     OnTouch(arguments, std::move(result));
+  } else if (method == "offset") {
+    OnOffset(arguments, std::move(result));
   } else {
     FT_LOG(Warn) << "Unimplemented method: " << method;
     result->NotImplemented();
@@ -204,6 +206,33 @@ void PlatformViewChannel::OnDispose(
   RemoveViewIfExists(*view_id);
 
   result->Success();
+}
+
+void PlatformViewChannel::OnOffset(
+    const EncodableValue* arguments,
+    std::unique_ptr<MethodResult<EncodableValue>>&& result) {
+  auto* map = std::get_if<EncodableMap>(arguments);
+  if (!map) {
+    result->Error("Invalid arguments");
+    return;
+  }
+
+  EncodableValueHolder<int> view_id(map, "id");
+  EncodableValueHolder<double> left(map, "left");
+  EncodableValueHolder<double> top(map, "top");
+
+  if (!view_id || !left || !top) {
+    result->Error("Invalid arguments");
+    return;
+  }
+  PlatformView* view = FindViewById(*view_id);
+  if (!view) {
+    result->Error("Can't find view id");
+    return;
+  }
+  view->Offset(*left, *top);
+
+  result->Success(*arguments);
 }
 
 void PlatformViewChannel::OnResize(
--- a/flutter/shell/platform/tizen/channels/platform_view_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_view_channel.cc
@@ -17,11 +17,13 @@ constexpr char kChannelName[] = "flutter/platform_views";
 
 }  // namespace
 
-PlatformViewChannel::PlatformViewChannel(BinaryMessenger* messenger)
+PlatformViewChannel::PlatformViewChannel(BinaryMessenger* messenger,
+                                         double pixel_ratio)
     : channel_(std::make_unique<MethodChannel<EncodableValue>>(
           messenger,
           kChannelName,
-          &StandardMethodCodec::GetInstance())) {
+          &StandardMethodCodec::GetInstance())),
+      pixel_ratio_(pixel_ratio) {
   channel_->SetMethodCallHandler(
       [this](const MethodCall<EncodableValue>& call,
              std::unique_ptr<MethodResult<EncodableValue>> result) {
@@ -149,8 +151,8 @@ void PlatformViewChannel::OnCreate(
     if (focused_view) {
       focused_view->SetFocus(false);
     }
-    PlatformView* view =
-        iter->second->Create(*view_id, *width, *height, byte_message);
+    PlatformView* view = iter->second->Create(
+        *view_id, *width * pixel_ratio_, *height * pixel_ratio_, byte_message);
     if (view) {
       views_[*view_id] = view;
       result->Success(EncodableValue(view->GetTextureId()));
@@ -230,7 +232,7 @@ void PlatformViewChannel::OnOffset(
     result->Error("Can't find view id");
     return;
   }
-  view->Offset(*left, *top);
+  view->Offset(*left * pixel_ratio_, *top * pixel_ratio_);
 
   result->Success();
 }
@@ -258,7 +260,7 @@ void PlatformViewChannel::OnResize(
     result->Error("Can't find view id");
     return;
   }
-  view->Resize(*width, *height);
+  view->Resize(*width * pixel_ratio_, *height * pixel_ratio_);
 
   result->Success(*arguments);
 }
@@ -285,10 +287,10 @@ void PlatformViewChannel::OnTouch(
 
   type = std::get<int>(event->at(0));
   button = std::get<int>(event->at(1));
-  x = std::get<double>(event->at(2));
-  y = std::get<double>(event->at(3));
-  dx = std::get<double>(event->at(4));
-  dy = std::get<double>(event->at(5));
+  x = std::get<double>(event->at(2)) * pixel_ratio_;
+  y = std::get<double>(event->at(3)) * pixel_ratio_;
+  dx = std::get<double>(event->at(4)) * pixel_ratio_;
+  dy = std::get<double>(event->at(5)) * pixel_ratio_;
 
   PlatformView* view = FindViewById(*view_id);
   if (!view) {
--- a/flutter/shell/platform/tizen/channels/platform_view_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_view_channel.cc
@@ -104,12 +104,12 @@ void PlatformViewChannel::HandleMethodCall(
     OnClearFocus(arguments, std::move(result));
   } else if (method == "dispose") {
     OnDispose(arguments, std::move(result));
+  } else if (method == "offset") {
+    OnOffset(arguments, std::move(result));
   } else if (method == "resize") {
     OnResize(arguments, std::move(result));
   } else if (method == "touch") {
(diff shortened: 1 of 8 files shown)


flutter/shell/platform/tizen/channels/platform_view_channel.cc
```diff
@@ -206,6 +208,33 @@ void PlatformViewChannel::OnDispose(
...
+    return;
+  }
+  PlatformView* view = FindViewById(*view_id);
+  if (!view) {
+    result->Error("Can't find view id");
+    return;
+  }
+  view->Offset(*left, *top);
+
+  result->Success(*arguments);
```
bbrto21, line 235: Why use `arguments` for success response?  is the dart side expecting to get it back?

flutter/shell/platform/tizen/channels/platform_view_channel.cc
```diff
@@ -108,6 +108,8 @@ void PlatformViewChannel::HandleMethodCall(
     OnResize(arguments, std::move(result));
   } else if (method == "touch") {
     OnTouch(arguments, std::move(result));
+  } else if (method == "offset") {
+    OnOffset(arguments, std::move(result));
```
JSUYA, line 112: I think it looks better to use them in the order they are defined.
"dispose"-> "offset"-> "resize"-> "touch"

flutter/shell/platform/tizen/channels/platform_view_channel.cc
JSUYA:
```java
        private void offset(@NonNull MethodCall call, @NonNull MethodChannel.Result result) {
          Map<String, Object> offsetArgs = call.arguments();
          try {
            handler.offset(
                (int) offsetArgs.get("id"),
                (double) offsetArgs.get("top"),
                (double) offsetArgs.get("left"));
            result.success(null);
          } catch (IllegalStateException exception) {
            result.error("error", detailedExceptionString(exception), null);
          }
        }
```
Android implementation is passing null.
probably, you don't need passing arguments.

JSUYA (commented):
Is there any way to test it? "offset" is still not called. (I tested with webview_flutter example)

flutter/shell/platform/tizen/channels/platform_view_channel.cc
Swanseo0: Changed to pass null. thank you!

flutter/shell/platform/tizen/channels/platform_view_channel.cc
Swanseo0: Modified. thank you.

Swanseo0:
> Is there any way to test it? "offset" is still not called. (I tested with webview_flutter example)

[[webview_flutter_tizen] Fix youtube player on TV_PROFILE ](https://github.com/flutter-tizen/plugins/pull/564)
Copy [flutter_platform_view.h](https://github.com/flutter-tizen/embedder/pull/33/files#diff-ad1e170127e27374d72dfc996d4cb162441b78bd18720ff7a1363a3f148afb3c) to `/flutter-tizen/flutter/bin/cache/artifacts/engine/tizen-common/public` and apply PR on the link above, and you can test it.

@JSUYA

flutter/shell/platform/tizen/channels/platform_view_channel.cc
bbrto21: Note: I think that when writing a `MethodResult` response, you should look at the corresponding Dart implementation.
In this case, maybe you should take a look [here](https://github.com/flutter-tizen/flutter-tizen/blob/e3fa9d40b4273e0bb3b78c50a5fbdfdf25b5f927/packages/flutter_tizen/lib/src/services/platform_views.dart#L240).

Thanks you for response.

bbrto21 (approved):
LGTM

JSUYA (approved):
For this patch to be merged, Also `flutter_webview_lwe` need to be modified.

flutter/shell/platform/tizen/channels/platform_view_channel.cc
```diff
@@ -147,8 +149,13 @@ void PlatformViewChannel::OnCreate(
     if (focused_view) {
       focused_view->SetFocus(false);
     }
-    PlatformView* view =
-        iter->second->Create(*view_id, *width, *height, byte_message);
+#ifdef TV_PROFILE
+    double pixel_ratio = 1.6;
+#else
+    double pixel_ratio = 1.0;
```
swift-kim, line 155: Why hard-coded?

flutter/shell/platform/tizen/public/flutter_platform_view.h
```diff
@@ -78,6 +79,7 @@ class PlatformViewFactory {
   virtual PlatformView* Create(int view_id,
                                double width,
                                double height,
+                               double pixel_ratio,
```
swift-kim, line 82: Shouldn't we pass physical pixel values in `Create`, `Offset`, and `Resize`? Logical sizes are values internally used by the framework (in Dart code), and it's the embedder's responsibility to process the values into raw pixels. See the external texture API (`flutter_texture_registrar.h`) for comparison.

flutter/shell/platform/tizen/channels/platform_view_channel.cc
Swanseo0: The reason why I hardcoded was that I did 1.6 with the fixed dpi value of 72 on the TV, and I didn't consider other pofiles yet, so the default value was 1.0.
However, I'll fix it in a way that calculates and uses it.

flutter/shell/platform/tizen/public/flutter_platform_view.h
Swanseo0: I'll remove the pixel_ratio from the argument again.
The reason is that the `platformViewChannel` cannot obtain the required dpi when calculating the pixel_ratio.

> Shouldn't we pass physical pixel values in Create, Offset, and Resize?

It looks good to pass the logical lengths from the dart code to `PlatformView`. `PlatformView` calculates and uses `pixel_ratio`.

swift-kim: > The reason is that the `platformViewChannel` cannot obtain the required dpi when calculating the pixel_ratio.

Why? Please, please discuss with your colleagues before making any decision.

> It looks good to pass the logical lengths from the dart code to `PlatformView`.

I have explained the reason for my suggestion. What's your rationale?

Swanseo0: > Why? Please, please discuss with your colleagues before making any decision.

I'm sorry. I'll discuss it with my colleague before I make a decision next time.
> I have explained the reason for my suggestion. What's your rationale?

we need `pixel_ratio` to convert `logical pixel` to `physical pixel`. By the way, in order to get `pixel_ratio` in flutter-tizen, `evas` or `evas_object` is required. These are variables that we have in `PlatformView(WebView)`, and these are variables that we can't get in `PlatformViewChannel`, so I think we should convert `logical pixel` to `physical pixel` in `PlatformView`.

flutter/shell/platform/tizen/flutter_tizen_engine.cc
```diff
@@ -376,10 +371,9 @@ void FlutterTizenEngine::SetupLocales() {
   // Convert the locale list to the locale pointer list that must be provided.
   std::vector<const FlutterLocale*> flutter_locale_list;
   flutter_locale_list.reserve(flutter_locales.size());
-  std::transform(
-      flutter_locales.begin(), flutter_locales.end(),
-      std::back_inserter(flutter_locale_list),
-      [](const auto& arg) -> const auto* { return &arg; });
+  std::transform(flutter_locales.begin(), flutter_locales.end(),
+                 std::back_inserter(flutter_locale_list),
+                 [](const auto& arg) -> const auto* { return &arg; });
```
swift-kim, line 376: The latest VS Code C++ extension uses clang-format-16 by default as I mentioned earlier in the messenger. You should manually specify the clang-format path with `/usr/bin/clang-format-15` in the VS Code settings. This will be fixed in the future.

flutter/shell/platform/tizen/flutter_tizen_view.cc
```diff
@@ -49,6 +49,20 @@ const std::vector<std::string> kBindableSystemKeys = {
...
+  // profile. A fixed DPI value (72) is used on TVs. See:
+  // https://docs.tizen.org/application/native/guides/ui/efl/multiple-screens
+#ifdef TV_PROFILE
+  double dpi = 72.0;
+#else
+  double dpi = static_cast<double>(view->GetDpi());
+#endif
+  double scale_factor = dpi / 90.0 * kProfileFactor;
+  double pixel_ratio = std::max(scale_factor, 1.0);
+  return pixel_ratio;
```
swift-kim, line 63: nit:
```suggestion
  return std::max(scale_factor, 1.0);
```

flutter/shell/platform/tizen/flutter_tizen_view.cc
```diff
@@ -375,16 +396,7 @@ void FlutterTizenView::SendWindowMetrics(int32_t left,
...
-    // profile. A fixed DPI value (72) is used on TVs. See:
-    // https://docs.tizen.org/application/native/guides/ui/efl/multiple-screens
-#ifdef TV_PROFILE
-    double dpi = 72.0;
-#else
-    double dpi = static_cast<double>(tizen_view_->GetDpi());
-#endif
-    double scale_factor = dpi / 90.0 * kProfileFactor;
-    computed_pixel_ratio = std::max(scale_factor, 1.0);
+    computed_pixel_ratio = ComputePixelRatio(tizen_view_.get());
```
swift-kim, line 399: The `computed_pixel_ratio` variable should not be necessary.
```suggestion
    pixel_ratio = ComputePixelRatio(tizen_view_.get());
```

swift-kim (commented):
Otherwise looks good to me.

flutter/shell/platform/tizen/public/flutter_platform_view.h
```diff
@@ -30,6 +30,7 @@ class PlatformView {
 
   virtual void Dispose() = 0;
 
+  virtual void Offset(double left, double top) = 0;
```
swift-kim, line 33: Could you provide a default implementation so that we don't break existing apps and plugins?

flutter/shell/platform/tizen/public/flutter_platform_view.h
```diff
@@ -30,6 +30,7 @@ class PlatformView {
 
   virtual void Dispose() = 0;
 
+  virtual void Offset(double left, double top){};
```
swift-kim, line 33:
```suggestion
  virtual void Offset(double left, double top) {}
```

Swanseo0: Modified. Thank you.

swift-kim (approved):

PR #34, opened 2023-05-08
[TizenAppControl] Add SetAutoRestart method

hjhun:
The app_control_set_auto_restart() is supported since tizen 7.0.
The application can set the auto restart option at runtime using the function.
To use the function, the application has to be signed using the platform level certificate.

### DIFF
--- a/flutter/shell/platform/tizen/channels/app_control.cc
+++ b/flutter/shell/platform/tizen/channels/app_control.cc
@@ -369,4 +369,12 @@ AppControlResult AppControl::Reply(AppControl* reply,
                                              result_e);
 }
 
+AppControlResult AppControl::SetAutoRestart(bool enabled) {
+  if (enabled) {
+    return app_control_set_auto_restart(handle_);
+  }
+
+  return app_control_unset_auto_restart();
+}
+
 }  // namespace flutter
--- a/tools/generate_sysroot.py
+++ b/tools/generate_sysroot.py
@@ -137,10 +137,10 @@ def generate_sysroot(sysroot: Path, api_version: float, arch: str, quiet=False):
   else:
     sys.exit('Unknown arch: ' + arch)
 
-  base_repo = 'http://download.tizen.org/snapshots/tizen/{}-base/latest/repos/standard/packages'.format(
-    api_version)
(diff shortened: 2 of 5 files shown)


tools/generate_sysroot.py
```diff
@@ -137,10 +137,10 @@ def generate_sysroot(sysroot: Path, api_version: float, arch: str, quiet=False):
   else:
     sys.exit('Unknown arch: ' + arch)
 
-  base_repo = 'http://download.tizen.org/snapshots/tizen/{}-base/latest/repos/standard/packages'.format(
-    api_version)
-  unified_repo = 'http://download.tizen.org/snapshots/tizen/{}-unified/latest/repos/standard/packages'.format(
-    api_version)
+  base_repo = 'http://download.tizen.org/snapshots/TIZEN/Tizen-{}/Tizen-{}-Base/latest/repos/standard/packages'.format(
```
swift-kim, line 140: What's the difference between the directories `TIZEN` and `tizen`? Is `tizen` deprecated?

flutter/shell/platform/tizen/channels/app_control.cc
```diff
@@ -369,4 +369,12 @@ AppControlResult AppControl::Reply(AppControl* reply,
                                              result_e);
 }
 
+AppControlResult AppControl::SetAutoRestart(bool enabled) {
+  if (enabled) {
+    return app_control_set_auto_restart(handle_);
```
swift-kim, line 374: Our target API level is Tizen 5.5. Tizen 7.0-specific API cannot be used in the embedder (especially if it's an internal API).

Can't you just call `app_control_set_auto_restart` using FFI in your app?

hjhun: The robot platform team wants to use this API to set/unset auto restart at runtime.
It's a requirement from Robot Platform team.
I'll discuss it with the developer again.
Thank you for reviewing this PR.

tools/generate_sysroot.py
hjhun: tizen directory was deprecated.

Currently, the build system(quickbuild) is uploading new binaries in 'Tizen' directory.

PR #35, opened 2023-05-10
Fix memory leak issue

xiaowei-guan:
std::unique_ptr::release, just releases the ownership of the managed object, the caller is responsible for deleting the object.


flutter/shell/platform/tizen/external_texture_surface_egl.cc
```diff
@@ -38,7 +38,6 @@ ExternalTextureSurfaceEGL::~ExternalTextureSurfaceEGL() {
   if (state_->gl_texture != 0) {
     glDeleteTextures(1, static_cast<GLuint*>(&state_->gl_texture));
   }
-  state_.release();
```
swift-kim, line 41: The same thing may be applied to `external_texture_surface_evas_gl.cc`.

swift-kim (approved):
Thank you

PR #36, opened 2023-05-16
[engine] Sync Flutter 3.10 source code

swift-kim:
Copy `embedder.h` and `client_wrapper` from the Flutter 3.10 engine source tree.


swift-kim:
I'm going to create two more PRs that depend on this PR. I'll merge this PR without approval if there's no review by next Monday. 

Please also review my recent PRs in this and other repos so that we can finish the migration to Flutter 3.10 quickly.

flutter/shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
```diff
@@ -13,16 +16,16 @@ class EncodableValue;
 
 template <typename T = EncodableValue>
 struct StreamHandlerError {
-  const std::string& error_code;
-  const std::string& error_message;
```
bbrto21, line 17: This problem is finally solved!

bbrto21 (approved):
LGTM

flutter/shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
swift-kim: Yes, the workaround you applied to the plugins can now be removed but this change will not break the plugins since there's no plugin that utilizes the `error_details` field.

PR #38, opened 2023-05-19
Update libcxx and other dependencies

swift-kim:
- Align libcxx, libcxxabi, dart-sdk, clang, and ninja versions with Flutter 3.10.
- Compile libcxx using C++20. See https://github.com/flutter/buildroot/commit/633907dfb90f6e8d97ad122aa66cf2d4207cda9b for details.
- Resolve deprecation warnings.
- Format the source code with the updated clang-format (clang-format-16).

### DIFF
--- a/flutter/third_party/accessibility/ax/ax_event_generator.h
+++ b/flutter/third_party/accessibility/ax/ax_event_generator.h
@@ -125,9 +125,14 @@ class AX_EXPORT AXEventGenerator : public AXTreeObserver {
     const EventParams& event_params;
   };
 
-  class AX_EXPORT Iterator
-      : public std::iterator<std::input_iterator_tag, TargetedEvent> {
+  class AX_EXPORT Iterator {
    public:
+    using iterator_category = std::input_iterator_tag;
+    using value_type = TargetedEvent;
+    using difference_type = std::ptrdiff_t;
+    using pointer = TargetedEvent*;
+    using reference = TargetedEvent&;
+
     Iterator(
         const std::map<AXNode*, std::set<EventParams>>& map,
         const std::map<AXNode*, std::set<EventParams>>::const_iterator& head);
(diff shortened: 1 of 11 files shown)


flutter/third_party/accessibility/ax/ax_event_generator.h
```diff
@@ -125,9 +125,14 @@ class AX_EXPORT AXEventGenerator : public AXTreeObserver {
     const EventParams& event_params;
   };
 
-  class AX_EXPORT Iterator
-      : public std::iterator<std::input_iterator_tag, TargetedEvent> {
+  class AX_EXPORT Iterator {
```
bbrto21, line 128: Could you please explain a bit more? I can't find the rationale for this change.

swift-kim: `std::iterator` is deprecated and generates a deprecation warning (which is treated as an error). I referred to the current [Chromium source code](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/ui/accessibility/ax_event_generator.h) (specifically [this commit](https://chromium.googlesource.com/chromium/src/+/c40a8c4a15034b783944bed298cc1151c4e822d5)) to minimize the use of `#pragma`.

c.f. I couldn't find a replacement for `std::codecvt_utf8` in `Utf8ToUtf32CodePoint`.

bbrto21: flutter upstream still use [std::iterator](https://github.com/flutter/engine/blob/52d05954815da75f492c94c70607dc4a8d6cd7a8/third_party/accessibility/ax/ax_event_generator.h#L129), are these differences okay?

bbrto21: > c.f. I couldn't find a replacement for std::codecvt_utf8 in Utf8ToUtf32CodePoint.

I think we need to write own one or it just be maintained until the standard provides an appropriate alternative.

bbrto21 (approved):

flutter/third_party/accessibility/ax/ax_event_generator.h
swift-kim: Should be no problem. They will eventually update their code too.

swift-kim:
@flutter-tizen/maintainers `gclient sync -D` must be run before you build the embedder locally after this change.

PR #43, opened 2023-08-22
Fix coredump issue during exit app

xiaowei-guan:
Need check whether vsync_waiter_ is a null pointer, because during stop engine process, may receive a request for vsync.


flutter/shell/platform/tizen/flutter_tizen_engine.h
```diff
@@ -272,6 +272,7 @@ class FlutterTizenEngine {
   std::unique_ptr<TizenRenderer> renderer_;
 
 #ifndef WEARABLE_PROFILE
+  std::mutex vsync_mutex_;
```
swift-kim, line 275: nit: A blank line may be added after this.

swift-kim (approved):
Thank you!

PR #45, opened 2023-09-07
Add LiveText.isLiveTextInputAvailable to unsupported method

JSUYA:
LiveText.isLiveTextInputAvailable is a method not supported in Tizen.

### DIFF
--- a/flutter/shell/platform/tizen/channels/platform_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_channel.cc
@@ -39,6 +39,8 @@ constexpr char kSetPreferredOrientationsMethod[] =
     "SystemChrome.setPreferredOrientations";
 constexpr char kSetSystemUIOverlayStyleMethod[] =
     "SystemChrome.setSystemUIOverlayStyle";
+constexpr char kIsLiveTextInputAvailable[] =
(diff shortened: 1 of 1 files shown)


flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -26,6 +26,10 @@ constexpr char kChannelName[] = "flutter/platform";
 constexpr char kGetClipboardDataMethod[] = "Clipboard.getData";
 constexpr char kSetClipboardDataMethod[] = "Clipboard.setData";
 constexpr char kClipboardHasStringsMethod[] = "Clipboard.hasStrings";
+
+constexpr char kIsLiveTextInputAvailable[] =
+    "LiveText.isLiveTextInputAvailable";
+
```
swift-kim, line 32: It should make more sense if this line comes after `kSetSystemUIOverlayStyleMethod` (since `kIsLiveTextInputAvailable` comes after `kSetSystemUIOverlayStyleMethod` at line 166).

JSUYA: done

swift-kim (approved):

PR #46, opened 2023-09-07
Add exit application method to system channel

JSUYA:
Add System.exitApplication and System.requestAppExit method.
System.exitApplication requests two types (required, cancelable).

https://github.com/flutter-tizen/embedder/issues/47

### DIFF
--- a/flutter/shell/platform/tizen/channels/platform_channel.cc
+++ b/flutter/shell/platform/tizen/channels/platform_channel.cc
@@ -14,6 +14,7 @@
 #ifdef COMMON_PROFILE
 #include "flutter/shell/platform/tizen/channels/tizen_shell.h"
 #endif
+#include "flutter/shell/platform/tizen/flutter_tizen_view.h"
 #include "flutter/shell/platform/tizen/logger.h"
 #include "flutter/shell/platform/tizen/tizen_view.h"
 #include "flutter/shell/platform/tizen/tizen_window.h"
@@ -29,6 +30,8 @@ constexpr char kSetClipboardDataMethod[] = "Clipboard.setData";
 constexpr char kClipboardHasStringsMethod[] = "Clipboard.hasStrings";
 constexpr char kExitApplicationMethod[] = "System.exitApplication";
 constexpr char kRequestAppExitMethod[] = "System.requestAppExit";
+constexpr char kInitializationCompleteMethod[] =
+    "System.initializationComplete";
 constexpr char kPlaySoundMethod[] = "SystemSound.play";
 constexpr char kHapticFeedbackVibrateMethod[] = "HapticFeedback.vibrate";
 constexpr char kSystemNavigatorPopMethod[] = "SystemNavigator.pop";
@@ -45,7 +48,6 @@ constexpr char kSetSystemUIOverlayStyleMethod[] =
 constexpr char kIsLiveTextInputAvailable[] =
     "LiveText.isLiveTextInputAvailable";
 
-constexpr char kExitCodeKey[] = "exitCode";
 constexpr char kExitTypeKey[] = "type";
 constexpr char kExitTypeCancelable[] = "cancelable";
 constexpr char kExitTypeRequired[] = "required";
@@ -165,23 +167,24 @@ void PlatformChannel::HandleMethodCall(
 
     const std::string& exit_type = iter->value.GetString();
     if (exit_type == kExitTypeRequired) {
-      ui_app_exit();
-    } else if (exit_type == kExitTypeCancelable) {
-      iter = document.FindMember(kExitCodeKey);
-      if (iter == document.MemberEnd() || !iter->value.IsInt()) {
-        result->Error(kExitRequestError, kInvalidExitRequestMessage);
-        return;
+      if (auto* view = dynamic_cast<FlutterTizenView*>(view_->GetView())) {
+        view->ExitApplication();
       }
-      int exit_code = iter->value.GetInt();
+    } else if (exit_type == kExitTypeCancelable) {
       rapidjson::Document document;
       document.SetObject();
-      RequestAppExit(exit_type, exit_code);
+      RequestAppExit(exit_type);
       document.AddMember(kExitResponseKey, kExitResponseCancel,
                          document.GetAllocator());
       result->Success(document);
     } else {
-      result->Error(kExitRequestError, "Invalid type." + exit_type);
+      result->Error(kExitRequestError, "Invalid type.");
     }
+  } else if (method == kInitializationCompleteMethod) {
+    if (auto* view = dynamic_cast<FlutterTizenView*>(view_->GetView())) {
+      view->SetInitializationComplete(true);
+    }
+    result->Success();
   } else if (method == kRestoreSystemUiOverlaysMethod) {
     RestoreSystemUiOverlays();
     result->Success();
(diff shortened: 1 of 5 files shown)


JSUYA:
System.initializationComplete method should be added.

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -138,6 +154,35 @@ void PlatformChannel::HandleMethodCall(
...
+    rapidjson::Value::ConstMemberIterator iter =
+        document.FindMember(kExitTypeKey);
+    if (iter == document.MemberEnd()) {
+      result->Error(kExitRequestError, kInvalidExitRequestMessage);
+      return;
+    }
+
+    const std::string& exit_type = iter->value.GetString();
+    if (exit_type == kExitTypeRequired) {
+      if (auto* view = dynamic_cast<FlutterTizenView*>(view_->GetView())) {
```
swift-kim, line 168: To avoid non-obvious dynamic casts, you could pass a pointer to `FlutterTizenView` instead of `TizenViewBase` when creating the `PlatformChannel`.

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -138,6 +154,35 @@ void PlatformChannel::HandleMethodCall(
...
+        view->ExitApplication();
+      }
```
swift-kim, line 170: No result (`result->Success()`) being sent in this case.

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -138,6 +154,35 @@ void PlatformChannel::HandleMethodCall(
     document.AddMember(rapidjson::Value(kValueKey, allocator),
                        rapidjson::Value(ClipboardHasStrings()), allocator);
     result->Success(document);
+  } else if (method == kExitApplicationMethod) {
+    const rapidjson::Value& document = *arguments;
+    rapidjson::Value::ConstMemberIterator iter =
+        document.FindMember(kExitTypeKey);
+    if (iter == document.MemberEnd()) {
+      result->Error(kExitRequestError, kInvalidExitRequestMessage);
```
swift-kim, line 162: nit: You might inline `kInvalidExitRequestMessage` for consistency. (See other occurrences of `result->Error()` in this file.)
```suggestion
      result->Error(kExitRequestError, "Invalid application exit request.");
```

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -138,6 +154,35 @@ void PlatformChannel::HandleMethodCall(
...
+    const std::string& exit_type = iter->value.GetString();
+    if (exit_type == kExitTypeRequired) {
+      if (auto* view = dynamic_cast<FlutterTizenView*>(view_->GetView())) {
+        view->ExitApplication();
+      }
+    } else if (exit_type == kExitTypeCancelable) {
+      rapidjson::Document document;
+      document.SetObject();
+      RequestAppExit(exit_type);
+      document.AddMember(kExitResponseKey, kExitResponseCancel,
```
swift-kim, line 175: This looks a bit weird. Let's change the order.
```suggestion
      RequestAppExit(exit_type);
      rapidjson::Document document;
      document.SetObject();
      document.AddMember(kExitResponseKey, kExitResponseCancel,
```

flutter/shell/platform/tizen/flutter_tizen_view.cc
```diff
@@ -145,6 +145,12 @@ void FlutterTizenView::Resize(int32_t width, int32_t height) {
   tizen_view_->SetGeometry(geometry);
 }
 
+void FlutterTizenView::ExitApplication() {
+  if (initialization_complete_) {
+    ui_app_exit();
+  }
```
swift-kim, line 151: Could you explain a bit more? According to the Linux implementation (`system_exit_application`),

- If `kExitTypeRequired` or **not** `app_initialization_complete`, `quit_application` immediately.
- Otherwise, send a cancelable `request_app_exit` request. In the response callback, if `kExitResponseExit`, run `quit_application`.

Your implementation doesn't seem to match the behavior. I'd recommend moving the `initialization_complete_` field to the `PlatformChannel` class.

flutter/shell/platform/tizen/channels/platform_channel.cc
JSUYA: Moved FlutterTizenView's code to PlatformChannel.

flutter/shell/platform/tizen/channels/platform_channel.cc
JSUYA: done

flutter/shell/platform/tizen/channels/platform_channel.cc
JSUYA: Removed kInvalidExitRequestMessage.

flutter/shell/platform/tizen/flutter_tizen_view.cc
JSUYA: I referred to window implementation, but I misunderstood it.
Moved codes to PlatformChannel and modified its behavior.
Please review again. Thanks

flutter/shell/platform/tizen/channels/platform_channel.cc
```diff
@@ -140,6 +153,33 @@ void PlatformChannel::HandleMethodCall(
...
+    }
+
+    const std::string& exit_type = iter->value.GetString();
+    rapidjson::Document document;
+    document.SetObject();
+    if (!initialization_complete_ || exit_type == kExitTypeRequired) {
+      ui_app_exit();
+      document.AddMember(kExitResponseKey, kExitResponseExit,
+                         document.GetAllocator());
+      result->Success(document);
```
swift-kim, line 171: Question: Is it okay to execute any code after the `ui_app_exit` call?

flutter/shell/platform/tizen/channels/platform_channel.h
```diff
@@ -40,11 +40,16 @@ class PlatformChannel {
...
   void RestoreSystemUiOverlays();
+  void RequestAppExit(const std::string exit_type);
+  void RequestAppExitSuccess(const rapidjson::Document* result);
   void SetEnabledSystemUiOverlays(const std::vector<std::string>& overlays);
   void SetPreferredOrientations(const std::vector<std::string>& orientations);
 
   std::unique_ptr<MethodChannel<rapidjson::Document>> channel_;
 
+  // Whether or not initialization is complete from the platform.
+  bool initialization_complete_ = false;
```
swift-kim, line 51: What do you mean by "from the platform"? Isn't the initialization completion notified by the framework?

JSUYA: Fixed

flutter/shell/platform/tizen/channels/platform_channel.cc
JSUYA: As far as I know, Embedder's Shutdown() is called when ui_app_exit() is called and terminate() of service_app_lifecycle_callback_s is called.

But I don't know if `result->Success()` is always guaranteed to be called because we don't know when the terminate() cb will be called.
However, when I checked with the log, the code after ui_app_exit() was called.

swift-kim: I'm not sure if the terminate callback is called directly by the `ui_app_exit` function. If it is called synchronously, the engine will shut down immediately and the engine's message channel and the instance of this class (`PlatformChannel`) will become invalid right after the `ui_app_exit` call. If you didn't see an error like "Invalid engine handle" or the app didn't crash (SIGABRT), this should not be the case (i.e. the terminate callback is not called synchronously). We still don't know what's done internally by `ui_app_exit`.

JSUYA:
Rebased patch

flutter/shell/platform/tizen/channels/platform_channel.cc
JSUYA: @swift-kim When we call ui_app_exit(), ui_app_lifecycle_callback_s.terminate() is called at the end of the main loop.
[The documentation](https://docs.tizen.org/iot/api/5.0/tizen-iot-headed/group__CAPI__APPLICATION__MODULE.html#ga3f00a008e3612d97f68042635203d792) recommends deleting resources here.

terminate() calls FlutterDesktopViewDestroy() and the associated destructors are called sequentially.
The method call to Success() is expected to be processed before that. There were no related error messages or logs.

Please let me know if there is anything else I need to consider regarding this patch.

swift-kim (approved):

PR #48, opened 2023-09-21
Add the input device channel

swift-kim:
This is a feature requested by our customer (3rd-party TV app developer).

To retrieve the name of the last keyboard device used:
```dart
final device = await MethodChannel('tizen/internal/inputdevice')
    .invokeMapMethod<String, dynamic>('getLastUsedKeyboard');
final name = device!['name'] as String;
```
For TV remote controls, the result is either of the following.
- IR remote control: `wt61p807 rc device`
- Smart remote control (bluetooth): `Smart Control [year]`
- Virtual keyboard: `ime` (updated when closing the keyboard)
- Virtual numpad: `virtual key device` (updated on pressing the Done button on Tizen 6.5 or higher, otherwise updated immediately when pressed numbers)

For example, if the developer wants to detect whether the user used a virtual numpad to enter a number, they can compare the string with `virtual key device`.

### DIFF
--- a/flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
+++ b/flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
@@ -357,7 +357,8 @@ void TizenWindowEcoreWl2::RegisterEventHandlers() {
             if (!handled) {
               self->view_delegate_->OnKey(
                   key_event->key, key_event->string, key_event->compose,
-                  key_event->modifiers, key_event->keycode, true);
+                  key_event->modifiers, key_event->keycode,
+                  ecore_device_name_get(key_event->dev), true);
             }
             return ECORE_CALLBACK_DONE;
           }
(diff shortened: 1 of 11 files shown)


flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -357,7 +357,8 @@ void TizenWindowEcoreWl2::RegisterEventHandlers() {
             if (!handled) {
               self->view_delegate_->OnKey(
                   key_event->key, key_event->string, key_event->compose,
-                  key_event->modifiers, key_event->keycode, true);
+                  key_event->modifiers, key_event->keycode,
+                  ecore_device_name_get(key_event->dev), true);
```
JSUYA, line 361: question: Doesn’t input-event by the input panel update the device name?

JSUYA (approved):

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
swift-kim: The resulting name is `ime` when using the input panel (software keyboard). The `ECORE_EVENT_KEY_DOWN` event is triggered only once (when the panel is being closed).

JSUYA:
@swift-kim Can I merge this PR? Is there any reason not to do it

Issue #47, opened 2023-09-08
Update to SystemChannel


JSUYA:
Changes to SystemChannel.
- [x] https://github.com/flutter/flutter/pull/96637
- [x] https://github.com/flutter/flutter/pull/121378
- [x] https://github.com/flutter/flutter/pull/126075
- [x] https://github.com/flutter/flutter/pull/122885

_Originally posted by @JSUYA in https://github.com/flutter-tizen/embedder/issues/44#issuecomment-1695337962_

PR #49, opened 2023-10-05
Integrate keyboard channel

JSUYA:
Update KeyEvent class to KeyboardChannel class
and add "flutter/keyboard" channel for getKeyboardState method.

issue: https://github.com/flutter-tizen/embedder/issues/47


JSUYA:
This should only work if the device is a hardware keyboard. So I need to input device information. if #48 is merged, I will add more condition to TODO.

JSUYA:
I tested with this code :
```dart
import 'package:flutter/services.dart';
... 
    HardwareKeyboard.instance.addHandler(handleKeyEvent);
...

  bool handleKeyEvent(KeyEvent event) {
    if (event is KeyDownEvent) {
      print('Key event received: $event');
    }
    return true;
  }
```
result:
```
[I] flutter: Key event received: KeyDownEvent#aa3d5(physicalKey: PhysicalKeyboardKey#583f3(usbHidUsage: "0x00070052", debugName: "Arrow Up"), logicalKey: LogicalKeyboardKey#88515(keyId: "0x100000304", keyLabel: "Arrow Up", debugName: "Arrow Up"), character: null, timeStamp: 0:04:04.774216)
[I] flutter: Key event received: KeyDownEvent#e3589(physicalKey: PhysicalKeyboardKey#4fbfe(usbHidUsage: "0x0007004f", debugName: "Arrow Right"), logicalKey: LogicalKeyboardKey#857a2(keyId: "0x100000303", keyLabel: "Arrow Right", debugName: "Arrow Right"), character: null, timeStamp: 0:04:05.342287)
```

JSUYA:
@swift-kim Actually I don't know the exact purpose of syncKeyboardState()(use case ????). So.., when I called `HardwareKeyboard.instance.syncKeyboardState();` after putting a few keys in the application, I only checked that the pressed key list was called normally in the method call. Is there anything else I need to check?

swift-kim:
@JSUYA I was able to reproduce the original error in https://github.com/flutter/flutter/issues/87391 in this way:

1. Write a simple app that has a TextField widget and run it on a mobile emulator.
2. Hot restart the app while pressing the `Shift` key on the hardware keyboard attached to the emulator. (Note: I run `flutter-tizen run` over SSH so I can basically press `Shift`+`R` and `Shift` simultaneously on two different keyboards. Maybe you can use a physical device instead of an emulator if you don't have this setup.)
3. The app throws an exception.

```sh
Restarted application in 575ms.
[I] platform_channel.cc: HandleMethodCall(165) > Unimplemented method: LiveText.isLiveTextInputAvailable

══╡ EXCEPTION CAUGHT BY SERVICES LIBRARY ╞══════════════════════════════════════════════════════════
The following assertion was thrown during a platform message callback:
A KeyUpEvent is dispatched, but the state shows that the physical key is not pressed. If this occurs
in real application, please report this bug to Flutter. If this occurs in unit tests, please ensure
that simulated events follow Flutter's event model as documented in `HardwareKeyboard`. This was the
event: KeyUpEvent#2a738(physicalKey: PhysicalKeyboardKey#ed430(usbHidUsage: "0x000700e1", debugName:
"Shift Left"), logicalKey: LogicalKeyboardKey#df100(keyId: "0x200000102", keyLabel: "Shift Left",
debugName: "Shift Left"), character: null, timeStamp: 0:12:06.166821)
'package:flutter/src/services/hardware_keyboard.dart':
Failed assertion: line 471 pos 16: '_pressedKeys.containsKey(event.physicalKey)'

Either the assertion indicates an error in the framework itself, or we should provide substantially
more information in this error message to help you determine and fix the underlying cause.
In either case, please report this assertion by filing a bug on GitHub:
  https://github.com/flutter/flutter/issues/new?template=2_bug.yml

When the exception was thrown, this was the stack:
#2      HardwareKeyboard._assertEventIsRegular.<anonymous closure> (package:flutter/src/services/hardware_keyboard.dart:471:16)
#3      HardwareKeyboard._assertEventIsRegular (package:flutter/src/services/hardware_keyboard.dart:482:6)
#4      HardwareKeyboard.handleKeyEvent (package:flutter/src/services/hardware_keyboard.dart:601:5)
#5      KeyEventManager.handleRawKeyMessage (package:flutter/src/services/hardware_keyboard.dart:1021:37)
#6      BasicMessageChannel.setMessageHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:212:49)
#7      _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:567:35)
#8      _invoke2 (dart:ui/hooks.dart:202:13)
#9      _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#10     _Channel.push (dart:ui/channel_buffers.dart:135:31)
#11     ChannelBuffers.push (dart:ui/channel_buffers.dart:331:17)
#12     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:736:22)
#13     _dispatchPlatformMessage (dart:ui/hooks.dart:114:31)
(elided 2 frames from class _AssertionError)
════════════════════════════════════════════════════════════════════════════════════════════════════
```

JSUYA:
> @JSUYA I was able to reproduce the original error in [flutter/flutter#87391](https://github.com/flutter/flutter/issues/87391) in this way:

Thank you for your detail. I will re-open when PR is ready.

flutter/shell/platform/tizen/channels/keyboard_channel.cc
```diff
@@ -0,0 +1,65 @@
...
+    if (!pressing_records_.empty()) {
+      EncodableMap map;
+      for (const auto& key : pressing_records_) {
+        EncodableValue physical_value(static_cast<int64_t>(key.first));
+        EncodableValue logical_value(static_cast<int64_t>(key.second));
+        map[physical_value] = logical_value;
+      }
+
+      result->Success(EncodableValue(map));
+      return;
```
swift-kim, line 45: `return` may be omitted for consistency.
```suggestion
      result->Success(EncodableValue(map));
```

flutter/shell/platform/tizen/channels/keyboard_channel.cc
```diff
@@ -0,0 +1,65 @@
...
+    } else {
+      result->Success();
```
swift-kim, line 47: Is null a valid response? I think you can just return an empty map (i.e. the `pressing_records_.empty` check should not be necessary).

flutter/shell/platform/tizen/channels/keyboard_channel.h
```diff
@@ -0,0 +1,41 @@
...
+ private:
+  // Called when a method is called on |channel_|;
+  void HandleMethodCall(
+      const flutter::MethodCall<EncodableValue>& method_call,
+      std::unique_ptr<flutter::MethodResult<EncodableValue>> result);
+
+  // The MethodChannel used for communication with the Flutter engine.
+  std::unique_ptr<flutter::MethodChannel<EncodableValue>> channel_;
+
+  std::map<uint64_t, uint64_t> pressing_records_;
```
swift-kim, line 36: `pressing_records_` are already being tracked by the `KeyEventChannel`. Can we reuse the values? Please let me know if you want to discuss how to redesign or refactor the code.

flutter/shell/platform/tizen/channels/keyboard_channel.cc
JSUYA: done

flutter/shell/platform/tizen/channels/keyboard_channel.h
JSUYA: I updated KeyEventChannel to KeyboardChannel class and added keyboard method channel code

swift-kim (approved):
Nicely done! Thanks.

Issue #26, opened 2023-02-22
Symbol files created by the CI are not compatible with flutter-tizen-gdb


swift-kim:
A Dwarf error is generated when loading embedder symbols in the `flutter-tizen-gdb` console because the [compiler](https://github.com/flutter-tizen/embedder/blob/a152671a2226bec8e84d70a06c0aa05226eb8bc7/.github/workflows/build.yml#L56) used by the CI is too new compared to the GDB server version (7.8.1).

```text
(gdb) add-symbol-file /home/swift/Git/plugins/packages/flutter_reactive_ble/example/symbols/tizen-5.5-arm/libflutter_tizen_wearable.so
add symbol table from file "/home/swift/Git/plugins/packages/flutter_reactive_ble/example/symbols/tizen-5.5-arm/libflutter_tizen_wearable.so"
(y or n) y
Reading symbols from /home/swift/Git/plugins/packages/flutter_reactive_ble/example/symbols/tizen-5.5-arm/libflutter_tizen_wearable.so...
Dwarf Error: Cannot handle DW_FORM_strx1 in DWARF reader [in module /home/swift/Git/plugins/packages/flutter_reactive_ble/example/symbols/tizen-5.5-arm/libflutter_tizen_wearable.so]
(No debugging symbols found in /home/swift/Git/plugins/packages/flutter_reactive_ble/example/symbols/tizen-5.5-arm/libflutter_tizen_wearable.so)
(gdb)
```

https://stackoverflow.com/questions/17958650/receiving-gdb-error-dwarf-error-cannot-handle-dw-form-flag-present-in-dwarf-re

PR #50, opened 2023-10-06
[ci] Update LLVM 12 to 15

JSUYA:
clang version in ubuntu-latest(github action) changed from 12 to 13.
At least version 13 or higher should be used.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,7 +52,7 @@ jobs:
         run: |
           src/tools/gn \
             --target-cpu ${{ matrix.arch }} \
-            --target-toolchain /usr/lib/llvm-12 \
+            --target-toolchain /usr/lib/llvm-15 \
             --target-dir build
           ninja -C src/out/build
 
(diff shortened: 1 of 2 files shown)


JSUYA:
https://github.com/actions/runner-images/blob/releases/ubuntu22/20231001/images/linux/Ubuntu2204-Readme.md

JSUYA:
clang-12 has been removed. (https://github.com/actions/runner-images/pull/8382)
I think https://github.com/flutter-tizen/embedder/actions/runs/6416820574/job/17450844279?pr=49 is reason of error occurs.

swift-kim:
The version was deliberately set to 12 due to [this issue](https://github.com/flutter-tizen/embedder/issues/26). Please change the runner image to `ubuntu-20.04` if LLVM 12 is not available on the latest image, or you may find another way to work around the issue.

Notes:
- You can use any LLVM version on your machine to build the embedder **locally**, but the binaries produced by **the CI** must support Tizen 5.5 (gdbserver 7.8.1).
- Tizen Studio's GCC 9.2 toolchain comes with a newer GDB (8.3.1) but it doesn't seem to be [compatible with](https://github.com/flutter-tizen/flutter-tizen/blob/7bb72513aab16cccbc2ed1f586a6b7fbd84e08ca/bin/flutter-tizen-gdb#L188) Tizen 5.5.

JSUYA:
> The version was deliberately set to 12 due to [this issue](https://github.com/flutter-tizen/embedder/issues/26). Please change the runner image to `ubuntu-20.04` if LLVM 12 is not available on the latest image, or you may find another way to work around the issue.

I will make a patch, change image version 22 to 20.

PR #51, opened 2023-10-06
[ci] Add clang-12 to build dependencies

JSUYA:
clang 12 has been removed from ubuntu-latest(22.04).
https://github.com/actions/runner-images/issues/8263

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y rpm2cpio cpio binutils-${{ matrix.triple }}
+          sudo apt-get install -y clang-12 rpm2cpio cpio binutils-${{ matrix.triple }}
 
       - name: Install depot_tools
         run: |


.github/workflows/build.yml
```diff
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
swift-kim, line 7: How about just installing `llvm-12` at line 34?

.github/workflows/build.yml
```diff
@@ -98,7 +98,7 @@ jobs:
 
   test:
     needs: build
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
swift-kim, line 101: You can leave other values as they are (`ubuntu-latest`) if there's no specific reason to change.

.github/workflows/build.yml
swift-kim: You may replace `llvm-12` with `clang-12` as recommended in the [README](https://github.com/flutter-tizen/embedder#prerequisites). It seems `clang` depends on `llvm` but not vice versa.

JSUYA: Thank you for your advice. I updated PR

swift-kim (approved):

PR #52, opened 2023-12-04
Implement FlutterDesktopViewGetResourceId

xiaowei-guan:
Replace surface id with resource id for fixing overlap issue on Tizen TV  6.0.

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen.cc
@@ -240,6 +240,11 @@ void* FlutterDesktopViewGetNativeHandle(FlutterDesktopViewRef view_ref) {
   return view->tizen_view()->GetNativeHandle();
 }
 
+uint32_t FlutterDesktopViewGetResourceId(FlutterDesktopViewRef view_ref) {
+  flutter::FlutterTizenView* view = ViewFromHandle(view_ref);
+  return view->tizen_view()->GetResourceId();
+}
+
 void FlutterDesktopViewResize(FlutterDesktopViewRef view,
                               int32_t width,
                               int32_t height) {
--- a/flutter/shell/platform/tizen/tizen_view_base.h
+++ b/flutter/shell/platform/tizen/tizen_view_base.h
@@ -35,13 +35,13 @@ class TizenViewBase {
   // Returns the geometry of the view.
   virtual TizenGeometry GetGeometry() = 0;
 
-  virtual uint32_t GetResourceId() = 0;
-
   // Set the geometry of the view.
   virtual bool SetGeometry(TizenGeometry geometry) = 0;
 
   // Returns the dpi of the screen.
   virtual int32_t GetDpi() = 0;
+
+  virtual uint32_t GetResourceId() = 0;
 
   virtual void UpdateFlutterCursor(const std::string& kind) = 0;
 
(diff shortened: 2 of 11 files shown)


JSUYA:
@xiaowei-guan Hi Can you please tell us the PR or codes that uses `FlutterDesktopViewGetResourceId()`?

xiaowei-guan:
> @xiaowei-guan Hi Can you please tell us the PR or codes that uses `FlutterDesktopViewGetResourceId()`? If used only in plugins, wouldn't it be better to use a method channel (window channel or platform(view) channel)?

Hello, this new API just for video_palyer_avplay plugin(https://github.com/xiaowei-guan/plugins/commit/f3cff83de6d0ffbc198ea1fc1230dd7ac91165bc)

flutter/shell/platform/tizen/flutter_tizen.cc
```diff
@@ -240,6 +240,11 @@ void* FlutterDesktopViewGetNativeHandle(FlutterDesktopViewRef view_ref) {
   return view->tizen_view()->GetNativeHandle();
 }
 
+uint32_t FlutterDesktopViewGetResourceId(FlutterDesktopViewRef view_ref) {
+  flutter::FlutterTizenView* view = ViewFromHandle(view_ref);
+  return view->tizen_view()->GetResourceId();
```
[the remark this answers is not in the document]
xiaowei-guan: This new API just for video_palyer_avplay plugin(https://github.com/xiaowei-guan/plugins/commit/f3cff83de6d0ffbc198ea1fc1230dd7ac91165bc)
Actually we can extend this api for Ecore_Evas:
```cpp
   Ecore_Wl2_Window* wl_window = ecore_evas_wayland2_window_get(ee);
```
I only implemented it for Ecore_wayland, because this third-party wanted to apply the new API as soon as possible, I will add the rest of the code later.

swift-kim (approved):

flutter/shell/platform/tizen/tizen_view_base.h
```diff
@@ -35,6 +35,8 @@ class TizenViewBase {
   // Returns the geometry of the view.
   virtual TizenGeometry GetGeometry() = 0;
 
+  virtual uint32_t GetResourceId() = 0;
+
```
swift-kim, line 39: Nothing should come between `GetGeometry` and `SetGeometry`. Please move `GetResourceId` to line 46 (after `GetDpi`).

xiaowei-guan: Fixed

xuelian-bai (approved):

JSUYA (approved):
FYI :Maybe below code could get the same result.
I think, we are already using ext api in embedder, so current code is also good

```cpp
static Eina_Bool window_show(void *data ,
                                        int type EINA_UNUSED,
                                        void *event ) {
  Ecore_Wl2_Event_Window_Show *ev =
      static_cast<Ecore_Wl2_Event_Window_Show *>(event);
  if (data) {
    *reinterpret_cast<uint32_t *>(data) =
        static_cast<unsigned int>(ev->data[0]);
  }

  return ECORE_CALLBACK_PASS_ON;
}
...
ecore_event_handler_add(ECORE_WL2_EVENT_WINDOW_SHOW, window_show,
                          &resource_id_);

```

PR #56, opened 2024-03-13
[tools] Update base_repo and unified_repo of generate_sysroot script

xiaowei-guan:
The snapshots website has been changed, need update to latest.


swift-kim (approved):
Thanks!

PR #57, opened 2024-03-19
[engine] Sync Flutter 3.19.3 source code

JSUYA:
Copy embedder.h and client_wrapper from the Flutter 3.19.3 engine source tree.

+) Update actions version
engine: https://github.com/flutter-tizen/engine/tree/flutter-3.19.3


JSUYA:
`No space left on device` error occurs in CI. I have updated both the github action versions of the embedder and engine.
The embedder no longer occurs error, but the engine still occurs a `No space left on device` error during build.

swift-kim:
It seems `gclient` is unnecessarily fetching the Fuchsia SDK which is not required for our build. Could you check what change was made to `DEPS` and see if you can skip these steps?

```
[P3354 08:25:43.963 storage.go:245 I] [fetch  7/12] Fetching: 1685.3/1685.3 MB (100%, 141.18 MB/s)
[P3354 08:25:43.963 client.go:1834 I] [fetch  7/12] Fetched fuchsia/sdk/core/linux-amd64:I-3hiLjX2DDy6mu22Q51o4ktda9zg0wZWdk-QolLicoC in 12.8s

...

[P3354 08:26:14.877 reader.go:424 I] [unzip  7/12] Extracting (   3 files left): 6422.3/6429.2 MB (100%, 250.16 MB/s)
[P3354 08:26:16.546 client.go:1915 E] [unzip  7/12] Failed to install fuchsia/sdk/core/linux-amd64 in "src/fuchsia/sdk/linux": write /__w/engine/engine/.cipd/pkgs/11/Hz7i0XtjszmS/obj/arm64-api-16/sysroot/dist/lib/asan/ld.so.1: no space left on device (and 6320 other errors)
```

JSUYA:
> It seems `gclient` is unnecessarily fetching the Fuchsia SDK which is not required for our build. Could you check what change was made to `DEPS` and see if you can skip these steps?

The size of fuchsia sdk has increased from 2.3gb to 6.4gb.
In this [recently added patch](https://github.com/flutter/engine/commit/1e199ce6239d191a7d200d83a21da72dab9ea4f2), the `download_fuchsia_deps` flag has been added. So I apply part of the patch and set it to False.

swift-kim:
There seems to be a defect in your [recent patch](https://github.com/flutter-tizen/engine/commit/21180fdec19fe4be76d50e6fa70339dbf830a9de). The `actions/checkout` action was unexpectedly removed from the `check-symbols.yml` workflow. The workflow passed once due to the nature of the `workflow_run` trigger, but may fail the next time.

When you create a new branch in the engine repo and add a patch to the branch, you can open a PR instead of force pushing the commit to the branch if you want to get a review (unless the patch is trivial and you want to fix the CI immediately).

JSUYA:
> > There seems to be a defect in your [recent patch](https://github.com/flutter-tizen/engine/commit/21180fdec19fe4be76d50e6fa70339dbf830a9de). The `actions/checkout` action was unexpectedly removed from the `check-symbols.yml` workflow. The workflow passed once due to the nature of the `workflow_run` trigger, but may fail the next time.
> 
> ```
> --- a/.github/workflows/check-symbols.yml
> +++ b/.github/workflows/check-symbols.yml
> @@ -13,9 +13,7 @@ jobs:
>      runs-on: ubuntu-latest
> 
>      steps:
> -      - uses: actions/checkout@v3
> -
> -      - uses: actions/checkout@v3
> +      - uses: actions/checkout@v4
> ```
> 
> The same action was being executed twice, so I removed one. Does this affect the results? At least when I tested it, the check-symbol seemed to work normally when the build was successful.

swift-kim:
@JSUYA They are not the same and yes, the change affects the result.

> At least when I tested it, the check-symbol seemed to work normally when the build was successful.

This is because you didn't change the default branch.

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

I changed the default branch and now the same workflow will fail if run again.

swift-kim (approved):

PR #59, opened 2024-04-01
[ci] Fix check-symbol action fail

JSUYA:
This change adds missing code from the 3.19 update.
https://github.com/flutter-tizen/embedder/pull/57

This PR is valid until https://github.com/flutter-tizen/embedder/pull/58 is applied.


swift-kim:
This change should not be necessary. We can merge #58 directly.

PR #61, opened 2024-04-04
Support device type for KeyEvent

JSUYA:


JSUYA:
```dart
        KeyboardListener(
              autofocus: true,
              focusNode: FocusNode(),
              onKeyEvent: (event) {
                print(event.deviceType.label);
                print(event);
              },
              child: ...
            ),
```
I tested with above code, but remote controller still returns keyboard.
When I checking with log from the engine's FlutterEngineSendKeyEvent(), passed device_type is does not seem to be the problem.
This needs to be investigated some more.

JSUYA:
https://github.com/flutter/flutter/issues/168886

PR #62, opened 2024-04-04
[ci] Add Tizen 6.0 artifacts

JSUYA:
And set the default version to 6.0.
This task is part of https://github.com/flutter-tizen/flutter-tizen/issues/526

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        api-version: [5.5, 6.5]
+        api-version: ['5.5', '6.0', '6.5']
         arch: [arm, arm64, x86]
         include:
           - arch: arm
@@ -43,12 +43,16 @@ jobs:
           gclient config --name=src --unmanaged https://github.com/${{ github.repository }}
           gclient sync -v --no-history --shallow
 
+      - name: Generate Tizen 6.0 sysroot
+        if: ${{ matrix.api-version == '6.0' }}
+        run: src/tools/generate_sysroot.py --api-version 6.0 --out src/sysroot-6.0
+
       - name: Generate Tizen 6.5 sysroot
-        if: ${{ matrix.api-version == 6.5 }}
+        if: ${{ matrix.api-version == '6.5' }}
         run: src/tools/generate_sysroot.py --api-version 6.5 --out src/sysroot-6.5
 
       - name: Build for Tizen 5.5
-        if: ${{ matrix.api-version == 5.5 }}
+        if: ${{ matrix.api-version == '5.5' }}
         run: |
           src/tools/gn \
             --target-cpu ${{ matrix.arch }} \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -58,7 +58,7 @@ The Flutter embedder for Tizen.
 ### Notes
 
 - To build an app (TPK) with the embedder generated in the above, copy the output artifacts (`libflutter_tizen*.so`) into the [flutter-tizen](https://github.com/flutter-tizen/flutter-tizen) tool's cached artifacts directory (`flutter/bin/cache/artifacts/engine`) and run `flutter-tizen run` or `flutter-tizen build tpk`.
-- To use the target device's `libstdc++.so` instead of the embedder's built-in libc++ (`third_party/libcxx`), provide the `--system-cxx` option to `tools/gn`. Tizen 5.5's `libstdc++.so` is not compatible with C++17 so you must use a sysroot for Tizen 6.0 or above.
+- To use the target device's `libstdc++.so` instead of the embedder's built-in libc++ (`third_party/libcxx`), provide the `--system-cxx` option to `tools/gn`. If Tizen version is 6.0 or above, you must use a sysroot.
```
swift-kim, line 61: It's actually
```suggestion
- To use the target device's `libstdc++.so` instead of the embedder's built-in libc++ (`third_party/libcxx`), provide the `--system-cxx` option to `tools/gn`. Tizen 5.5's `libstdc++.so` is not compatible with C++17, so you cannot use this option for Tizen 5.5.
```
or you can simply remove the sentence (not very important information).

JSUYA: I have excluded code that changes the default version from this patch.
I will remove this sentence when excluding version 5.5 in the future.

.github/workflows/build.yml
```diff
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        api-version: [5.5, 6.5]
+        api-version: ['5.5', '6.0', '6.5']
```
swift-kim, line 11: Why not removing '5.5'?

JSUYA: I just thought I had to do it in order
add 6.0 artifact (embedder) -> Change default version 5.5 to 6.0 (flutter-tizen) -> Remove 5.5 artifact(embedder)

.github/workflows/build.yml
```diff
@@ -87,7 +102,7 @@ jobs:
           if-no-files-found: error
 
       - uses: actions/upload-artifact@v4
-        if: ${{ matrix.arch == 'arm' && matrix.api-version == 5.5 }}
+        if: ${{ matrix.arch == 'arm' && matrix.api-version == 6.0 }}
```
swift-kim, line 105: arm and 6.0 are both strings but the usage is not consistent.

```suggestion
        if: ${{ matrix.arch == 'arm' && matrix.api-version == '6.0' }}
```

swift-kim (approved):

.github/workflows/build.yml
JSUYA: Thank you for check. I also modified other lines.

PR #63, opened 2024-04-17
Add 8.0 to artifact creation version

JSUYA:
`cbhm` library has been deprecated in Tizen 8.0.
So, In common profile build was fail.
Supports creation of 8.0 artifact excluding `cbhm`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        api-version: [5.5, 6.5]
+        api-version: [5.5, 6.5, 8.0]
         arch: [arm, arm64, x86]
         include:
           - arch: arm
@@ -47,6 +47,10 @@ jobs:
         if: ${{ matrix.api-version == 6.5 }}
         run: src/tools/generate_sysroot.py --api-version 6.5 --out src/sysroot-6.5
 
+      - name: Generate Tizen 8.0 sysroot
+        if: ${{ matrix.api-version == 8 }}
+        run: src/tools/generate_sysroot.py --api-version 8.0 --out src/sysroot-8.0
+
       - name: Build for Tizen 5.5
         if: ${{ matrix.api-version == 5.5 }}
         run: |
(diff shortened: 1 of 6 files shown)


JSUYA:
https://github.com/flutter-tizen/flutter-tizen/issues/538

.github/workflows/build.yml
```diff
@@ -47,6 +47,10 @@ jobs:
         if: ${{ matrix.api-version == 6.5 }}
         run: src/tools/generate_sysroot.py --api-version 6.5 --out src/sysroot-6.5
 
+      - name: Generate Tizen 8.0 sysroot
+        if: ${{ matrix.api-version == 8 }}
```
swift-kim, line 51:
```suggestion
        if: ${{ matrix.api-version == 8.0 }}
```
Note that the official name is always "Tizen 8.0" but not "Tizen 8".

.github/workflows/build.yml
```diff
@@ -66,6 +70,17 @@ jobs:
             --api-version 6.5 --system-cxx \
             --target-dir build
           ninja -C src/out/build
+          
+      - name: Build for Tizen 8.0
```
swift-kim, line 74: I warned the Tizen UI framework team about the backward incompatibility but they proceeded with the API removal without providing any alternative API.

We need to inform the users that apps built for previous versions of Tizen (e.g. Tizen 6.5) are not compatible with Tizen 8.0+ devices. They have to upload two copies of TPK to the app store to support both <8.0 and 8.0+. 

Alternatively we can simply remove the functionality (clipboard support) or use `dlopen`.

.github/workflows/build.yml
```diff
@@ -66,6 +70,17 @@ jobs:
...
+        if: ${{ matrix.api-version == 8 }}
+        run: |
+          src/tools/gn \
+            --target-cpu ${{ matrix.arch }} \
+            --target-toolchain /usr/lib/llvm-12 \
+            --target-sysroot src/sysroot-8.0/${{ matrix.arch }} \
+            --api-version 8.0 --system-cxx \
```
swift-kim, line 81: This will effectively remove NUI support from the 8.0 embedder (see line 181 in `tizen/BUILD.gn`). If we're to compare more than two API versions, it would be better to create a dedicated command line option for each build flavor, e.g. `--[no]-nui`, `--[no]-cbhm`, because it's not quite easy to compare numbers in GN. Let me help if you're not familiar with the build system.

.github/workflows/build.yml
JSUYA: I will remove the existing cbhm code and make a PR again using the ecore_wl2 API. https://github.com/flutter-tizen/embedder/pull/64
Perhaps I expect copy&paste to be supported in all versions and TV and common profiles.

.github/workflows/build.yml
JSUYA: https://github.com/flutter-tizen/embedder/pull/62 I modified this patch to recognize the version as a string.

JSUYA:
I will re-process(?) with a new PR.

PR #66, opened 2024-04-25
Change minimum Tizen version 5.5 to 6.0

JSUYA:
- Remove Tizen 5.5 support.
- Remove Wearable support.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -58,7 +58,6 @@ jobs:
             --target-cpu ${{ matrix.arch }} \
             --target-toolchain /usr/lib/llvm-12 \
             --target-sysroot src/sysroot-6.0/${{ matrix.arch }} \
-            --system-cxx \
             --target-dir build
           ninja -C src/out/build
 
@@ -69,7 +68,7 @@ jobs:
             --target-cpu ${{ matrix.arch }} \
             --target-toolchain /usr/lib/llvm-12 \
             --target-sysroot src/sysroot-6.5/${{ matrix.arch }} \
-            --api-version 6.5 --system-cxx \
+            --api-version 6.5 \
             --target-dir build
           ninja -C src/out/build
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
 
     strategy:
       matrix:
-        api-version: ['5.5', '6.0', '6.5']
+        api-version: ['6.0', '6.5']
         arch: [arm, arm64, x86]
         include:
           - arch: arm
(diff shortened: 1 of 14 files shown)


JSUYA:
Hi @swift-kim I have a question.
This PR's build test is failing. I tried cleaning local git and building with gclient sync, but i can't reproduce the build failure. Even in my repo where this patch was pushed builds well.
(https://github.com/JSUYA/embedder/actions/runs/8889817049/job/24408777062)

At download.tizen.org, there is a version difference between 5.5 and 6.5 libstdc++, but between 6.0 and 6.5 is not different.
As far as I know, the --system-cxx command should work normally in 6.0 as well.
I haven't found the reason for this. Do you know a solution to this problem?

swift-kim:
Everything looks fine but I think you need to clear the GitHub Actions cache (or do not delete `Generate Tizen 6.0 sysroot` and `--target-sysroot`).

The gclient sync hook should have generated a Tizen 5.5 sysroot in the default location (`src/sysroot`) and the files are being cached wrongly.

.github/workflows/build.yml
```diff
@@ -43,31 +43,17 @@ jobs:
...
-
       - name: Build for Tizen 6.0
         if: ${{ matrix.api-version == '6.0' }}
         run: |
           src/tools/gn \
             --target-cpu ${{ matrix.arch }} \
             --target-toolchain /usr/lib/llvm-12 \
-            --target-sysroot src/sysroot-6.0/${{ matrix.arch }} \
-            --api-version 6.0 --system-cxx \
+            --system-cxx \
```
swift-kim, line 56: We could make `--system-cxx` default.

Please update `BUILDCONFIG.gn` as well as `tools/gn`. You can also remove the note in README.

flutter/shell/platform/tizen/tizen_evas_gl_helper.h
```diff
@@ -1,5 +1,5 @@
 // This file is a copy of Tizen's Evas_GL_GLES2_Helpers.h file.
-// https://docs.tizen.org/application/native/api/wearable/5.5/group__Evas__GL__GLES2__Helpers.html
+// https://docs.tizen.org/application/native/api/iot-headed/6.0/group__Evas__GL__GLES2__Helpers.html
```
swift-kim, line 2: Are the two files identical?

JSUYA: Yes, as far as I know it is the same

JSUYA:
> Everything looks fine but I think you need to clear the GitHub Actions cache (or do not delete `Generate Tizen 6.0 sysroot` and `--target-sysroot`).

Thanks. I will add `Generate Tizen 6.0 sysroot`.
Actually, I don't know if the cache for the default version worked well. (Part cached with sysroot-5.5)

> The gclient sync hook should have generated a Tizen 5.5 sysroot in the default location (`src/sysroot`) and the files are being cached wrongly.

Because I changed the default version of generate_sysroot.py to 6.0, Tizen 6.0 sysroot is generated in src/sysroot, so I think the sysroot generated with gclient sync + hook is Tizen 6.0 sysroot. 
Adding the `Generate Tizen 6.0 sysroot` job will solve the problem, but I don't know why src/sysroot is not working properly.

swift-kim:
Because the sysroot is not generated again (overwritten) if already present (i.e. cached).

Could you please refrain from force-pushing changes? It makes it difficult to track changes and thus is not good for reviewability.

.github/workflows/build.yml
swift-kim: Right. I think the option name should be `--no-system-cxx`.

JSUYA:
rebased pr

tools/gn
```diff
@@ -75,7 +75,7 @@ def to_gn_args(args):
 
   gn_args['clang_version'] = get_clang_version(args.target_toolchain)
   gn_args['api_version'] = args.api_version
-  gn_args['use_system_cxx'] = args.system_cxx
+  gn_args['no_system_cxx'] = args.no_system_cxx
```
swift-kim, line 78: Please do not change the config name (I hate [double negation](https://dart.dev/effective-dart/design#prefer-the-positive-name-for-a-boolean-property-or-variable)). I was only referring to the command line option name in the above comment.

```suggestion
  gn_args['use_system_cxx'] = not args.no_system_cxx
```

JSUYA: done

JSUYA:
@swift-kim Could you please approve?, if there is no more code that needs to be fixed. It seems to applied most of the reviews.

swift-kim (approved):

Issue #73, opened 2024-11-13
[a11y] elm_atspi_bridge_utils_say() does not work


JSUYA:
When trying to speak through `SemanticsService.announce()` for accessibility work, it internally calls the elm_atspi_bridge_utils_say() API to speak.

```
SemanticsService.announce("Speech", TextDirection.ltr);
```

```
E/EFL     (12574): elementary<12574> ../src/lib/elementary/elm_atspi_bridge.c:8196 elm_atspi_bridge_utils_say() AT-SPI: Atspi bridge is not enabled.
```

However, since TV profile does not use elm_win, the elm API elm_atspi_bridge_utils_say() does not work. Therefore, it must be updated to use the tts engine directly.

PR #75, opened 2024-12-03
[a11y] Support for "announce" of AccessibilityChannel

JSUYA:
elm_atspi_bridge_utils_say() does not work unless user build it with c++ native.
So SemanticsService.announce() does not work properly on TV devices.
Therefore, send messages directly to the Tizen screen-reader using eldbus instead of elm_atspi_bridge_utils_say().


JSUYA:
https://github.com/flutter-tizen/embedder/issues/73

xiaowei-guan (approved):

Issue #8, opened 2022-04-25
[a11y] ATK assertion failure when screen reader is enabled


swift-kim:
I see a lot of these error messages when running the Gallery app and moving widget focus on TV. Is this related to the recently added screen reader support?

```
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:41:58.122: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:42:04.997: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:42:05.677: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:42:06.067: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:42:07.470: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): (Runner.dll:14825): Atk-CRITICAL **: 14:42:10.939: atk_text_get_character_count: assertion 'ATK_IS_TEXT (text)' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): (Runner.dll:14825): Atk-CRITICAL **: 14:42:10.956: atk_text_get_text: assertion 'ATK_IS_TEXT (text)' failed
E/ConsoleMessage(14825):
E/ConsoleMessage(14825): ** (Runner.dll:14825): CRITICAL **: 14:42:19.624: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
E/ConsoleMessage(14825):
[...]

PR #76, opened 2024-12-04
[a11y] If there is no action, return immediately

JSUYA:
The index of DoAction is the number of the action that can be called.
If there is no action that can be called in atk, the index is passed as -1.
Currently, our AXPlatformNode only supports default actions, so we need to distinguish between -1 and numbers greater than 0.
In Tizen, the DoActionName function in atk causes this problem.
Since at-spi2-atk in tizen is not under our management scope, we temporarily return immediately when it is -1 to prevent unnecessary error messages.

error message
```
[E] ** (Runner.dll:10005): CRITICAL **: 20:27:38.451: gboolean ui::(anonymous namespace)::atk_action::DoAction(AtkAction *, gint): assertion '!index' failed
[E] [0]: /usr/lib/libglog-runtime-env.so+0x78a (g_log_default_handler+0xb9) [0xb618578a]
[E] [1]: /lib/libglib-2.0.so.0+0x5053e (g_logv+0x149) [0xb5a6053e]
[E] [2]: /lib/libglib-2.0.so.0+0x50686 (g_log+0x25) [0xb5a60686]
[3]: /lib/libglib-2.0.so.0+0x50832 (g_return_if_fail_warning+0x11) [0xb5a60832]

```


JSUYA:
https://github.com/flutter-tizen/embedder/issues/8

swift-kim (approved):

PR #85, opened 2025-04-16
Add setDirection method to platform view channel

seungsoo47:


JSUYA:
Can you share some example code where and how this method is used? It seems that setLayoutDirection is only used in Android views and iOS UIViews.

seungsoo47:
> Can you share some example code where and how this method is used? It seems that setLayoutDirection is only used in Android views and iOS UIViews.

Yes, you are correct.
The 'setDirection' method channel is used to set the text direction (LTR or RTL) and is passed from the setLayoutDirection() function.
The setLayoutDirection() function is called from didChangeDependencies() and didUpdateWidget() in platform_view.dart.
This method channel has been defined and implemented on both the Flutter and native sides quite a while ago.
(However, it's unclear why this method channel was missing in our embedder.)
In this patch, I have added only the part that calls the method channel.

Additionally, based on the current check, there doesn't seem to be a direct API to call setDirection.
If you'd like, it might be helpful to directly modify didChangeDependencies() and didUpdateWidget() for testing.
However, it seems there are bugs in these two functions, so it might not be an easy task.

JSUYA (approved):

Issue #55, opened 2024-02-23
Scaling factor


adamdama:
We are experimenting with delivering Flutter apps to Tizen based TVs. During our exploration we found that the app is being scaled up by a factor of 1.6 which seems to be coming from this code:
https://github.com/flutter-tizen/embedder/blob/c382987e93537a3be6b0c11eff4076ce4af6980b/flutter/shell/platform/tizen/flutter_tizen_view.cc#L61

This is making our app elements look too big even though the window size is correct at 1920 x 1080.

Is there a recommended approach to adjusting this scale in the Flutter app globally? Or do we need to adjust the widget and height of every widget when deployed to Tizen?

swift-kim:
As far as I remember, the DPI came from the TV UI design guideline, so the scale factor (pixel ratio) has a constant value regardless of the screen size. You may find the widgets too big if you have a large screen TV. All other built-in apps should follow the same design guidelines and therefore have the same look and feel. Maybe I need to check out the latest guidelines again.

[This issue](https://github.com/flutter/flutter/issues/32115) and the workaround suggested by ardera may help, but it doesn't seem to work as desired in all cases. The easiest way would be to build the embedder by yourself by adjusting the hardcoded value. You will need a Linux machine or WSL, or you can even use the [GitHub Actions](https://github.com/flutter-tizen/embedder/blob/master/.github/workflows/build.yml) CI if you don't want to set up the build environment locally. Otherwise we can consider adding a dedicated API to the [C# embedding](https://github.com/flutter-tizen/flutter-tizen/blob/master/embedding/csharp/Tizen.Flutter.Embedding/FlutterApplication.cs) but the implementation may take some time.

PR #86, opened 2025-04-23
Add user-defined pixel ratio

JSUYA:
Previously, we used values calculated through Base scale calculation using DPI and DeviceFactor used in Tizen.
Since it follows Tizen UX, it can show the same look as other apps on TV. (https://docs.tizen.org/application/native/guides/ui/efl/multiple-screens/)
However, when running an app written for another platform on Tizen, the scale may appear different.
So I add user_pixel_ratio to WindowProperty to allow the user to adjust this scale.


JSUYA:
+) @myroot

JSUYA:
https://github.com/flutter-tizen/embedder/issues/55

xiaowei-guan (approved):

PR #87, opened 2025-04-24
Add Twitter/WebSearch Keyboard

seungsoo47:
Add Twitter/WebSearch Keyboard layout.
Note that Tizen IMF only supports '@' for Twitter TextInputType and '/' for WebSearch TextInputType like Android.

https://github.com/flutter/flutter/pull/161025


JSUYA (approved):
LGTM

PR #97, opened 2025-05-29
Drop use of volatile with g_once_*()

JSUYA:
The variables here are only marked as volatile because they’re arguments to `g_once_*()`.

Related chromium commit:
https://chromium-review.googlesource.com/c/chromium/src/+/3960017
Related glib commit:
https://github.com/GNOME/glib/commit/fab561f8d0


JSUYA:
This problem was discovered during a pre-check of Tizen 10.0 support.

PR #104, opened 2025-06-25
Implement merged UI and platform thread

seungsoo47:


seungsoo47:
As discussed offline, we can merge this patch after this feature is enabled on Linux.

flutter/shell/platform/tizen/flutter_tizen_engine.cc
```diff
@@ -177,6 +177,14 @@ bool FlutterTizenEngine::RunEngine() {
     custom_task_runners.render_task_runner = &render_task_runner;
   }
 
+  if (project_->ui_thread_policy() !=
+      FlutterUIThreadPolicy::kRunOnSeparateThread) {
+    FT_LOG(Info) << "Running with merged platform and UI threads.";
+    custom_task_runners.ui_task_runner = &platform_task_runner;
+  } else {
+    FT_LOG(Info) << "Running with unmerged platform and UI threads.";
```
JSUYA: I don't think it's necessary to always output the info log. If default value is explicitly stated in the documentation, I believe it's sufficient to output logs only when something is configured.

Issue #106, opened 2025-07-01
Pressing the back button(remote control) in the dialog sends “SystemNavigator.pop” to the Platform channel


JSUYA:
flutter-tizen exits the app by calling ui_app_exit() when the "SystemNavigator.pop" method is called from the embedder's platform channel.

When "SystemNavigator.pop" is called, it means that the navigation inside the app no longer has anything to pop().
However, "SystemNavigator.pop" is also called for Dialogs.
When the dialog is called from the last navigator in the app, we want the dialog to close when the back button is pressed.
However, "SystemNavigator.pop" is called through the Platform channel.

We may need to review the scenario of closing the app further.

related issue: https://github.com/flutter-tizen/flutter-tizen/issues/640
example code

```dart
import 'package:flutter/material.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  void _showDialog() {
    showDialog(
      context: context,
      builder: (BuildContext context) {
        return AlertDialog(
          title: const Text('Simple Dialog'),
          content: const Text(
            'This is a simple dialog shown from the floating button.',
          ),
          actions: <Widget>[
            TextButton(
              child: const Text('Close'),
              onPressed: () {
                Navigator.of(context).pop();
              },
            ),
          ],
        );
      },
    );
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[const Text('Click FloatingActionButton.')],
        ),
      ),
      floatingActionButton: Row(
        mainAxisAlignment: MainAxisAlignment.end,
        children: [
          FloatingActionButton(
            onPressed: _showDialog,
[...]

PR #107, opened 2025-07-03
Fix back-key handling

JSUYA:
The backkey performs the function of going back the prev-page(Navigator) by calling popRoute when pressed.
This is because the back key is a system-level key and other platforms behave the same.
however, in the dialog case of flutter, there is a case where navigator.pop() is occured on key-down.
Therefore, when the key-down event is handled for the Back key, the up event is also ‘handled’.

related issue: https://github.com/flutter-tizen/embedder/issues/106

### DIFF
--- a/flutter/shell/platform/tizen/channels/keyboard_channel.cc
+++ b/flutter/shell/platform/tizen/channels/keyboard_channel.cc
@@ -36,8 +36,6 @@ constexpr char kKeyDown[] = "keydown";
 constexpr char kGtkToolkit[] = "gtk";
 constexpr char kLinuxKeyMap[] = "linux";
 
-constexpr char kBackKey[] = "XF86Back";
-
 constexpr size_t kMaxPendingEvents = 1000;
 
 uint32_t Utf8ToUtf32CodePoint(const char* utf8) {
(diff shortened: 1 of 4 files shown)


flutter/shell/platform/tizen/channels/keyboard_channel.cc
```diff
@@ -189,11 +191,25 @@ void KeyboardChannel::SendChannelEvent(const char* key,
...
+                 const uint8_t* reply, size_t reply_size) {
         if (reply != nullptr) {
           std::unique_ptr<rapidjson::Document> decoded =
               JsonMessageCodec::GetInstance().DecodeMessage(reply, reply_size);
           bool handled = (*decoded)[kHandledKey].GetBool();
+          // If System's back key is handled in key-down, it should be
+          // handled so that "popRoute" is not called in key-up.
+          if (symbol == kBackKey) {
+            if (handled && is_down) {
+              this->backkey_handled_ = true;
```
swift-kim, line 204: `flutter_tizen_view.cc` has the back key handling logic as you might already know. Is that not enough or is it not the right place to handle this case?

swift-kim: No, I don't understand the exact issue but I was just wondering why the code was split into two different files. I don't mind if that makes it more readable for you.

JSUYA: I moved the handling code for backkey to flutter_tizen_view.cc.

seungsoo47 (approved):

Issue #111, opened 2025-08-22
Support mouse pointer/touch disable


JSUYA:
https://developer.samsung.com/smarttv/develop/guides/user-interaction/mouse.html
There is guide on `pointing-device-support` and `floating.navigation` in the design guide for Tizen TV WebApp.
```
<tizen:setting pointing-device-support='enable' />
<tizen:metadata key="http://samsung.com/tv/metadata/floating.navigation" value="false"/>
```
Users can configure the behavior of the Cursor, Floating menu (right-click mouse) and D-PAD according to these options.

We couldn't support this feature using the existing Native APIs. However, I have confirmed that it can be supported using the implementation way of [EnableCursor()](https://github.com/flutter-tizen/embedder/blob/master/flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc#L166).
(We found `Mouse/Touch_Pointer_Support`, `Mouse/Touch_Pointer_Not_Allow` APIs.)

The guide is saying the toast popup to be displayed when an event occurs, but if the mouse is disabled, the ecore_wl2 event does not occur, so it is not known when the event will occur. We need a solution for this.

Related issue: https://github.com/flutter-tizen/flutter-tizen/issues/579

PR #112, opened 2025-08-22
Support pointing device setting

JSUYA:
related issue: https://github.com/flutter-tizen/embedder/issues/111
tools: https://github.com/flutter-tizen/flutter-tizen/pull/660

### DIFF
--- a/flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
+++ b/flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
@@ -97,8 +97,11 @@ std::string GetPreferenceKey(bool clear_exist_key) {
   }
 
   std::string app_id = id;
+  free(id);
+
   std::ostringstream boot_time_buffer;
   boot_time_buffer << boot_time;
+
   std::string preference_key =
       std::string(kSysPointingDeviceSupportToastSharedPreferenceKey) + "/" +
       app_id + "/" + boot_time_buffer.str();


JSUYA:
In the web app, Toast popup output is not repeated until the device is rebooted.
This implementation needs to be added.

JSUYA:
> In the web app, Toast popup output is not repeated until the device is rebooted.
> This implementation needs to be added.

I calculated device booting time and implemented it using tizen preference.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -223,6 +321,82 @@ void TizenWindowEcoreWl2::EnableCursor() {
...
+  int (*Mouse_Pointer_Support)(int type, void* ecore_wl2_win);
+  *(void**)(&Mouse_Pointer_Support) = dlsym(handle, "Mouse_Pointer_Support");
+
+  if (!Mouse_Pointer_Support) {
+    FT_LOG(Error) << "Could not load symbols from the library.";
+    dlclose(handle);
+    return;
+  }
+
+  Mouse_Pointer_Support(pointing_device_support_, ecore_wl2_window_);
```
seungsoo47, line 346: I am sorry, I can't find Mouse_Pointer_Support on my TV.
So, I have a question. The 'pointing_device_support_' variable is a boolean, but 'Mouse_Pointer_Support' requires an int type.
Is this okay?

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -273,6 +447,32 @@ void TizenWindowEcoreWl2::RegisterEventHandlers() {
...
+          } else if (!self->floating_menu_support_) {
+            self->SetFloatingMenuSupport();
+            if (!shown) {
+              self->ShowUnsupportedToast();
+              SetPointingDevicePreference();
+            }
+          }
+
+          self->show_unsupported_toast_ = true;
+          return ECORE_CALLBACK_PASS_ON;
```
seungsoo47, line 472:
```suggestion
          bool shown = GetPointingDeviceToastPreference();
          if (!shown) {
            // Toast popup should be called first to set up D-PAD.
            self->ShowUnsupportedToast();
            SetPointingDevicePreference();
          }
          self->show_unsupported_toast_ = true;

          if (!self->pointing_device_support_) {
            self->SetPointingDeviceSupport();
          } else if (!self->floating_menu_support_) {
            self->SetFloatingMenuSupport();
          }
          return ECORE_CALLBACK_PASS_ON;
```
I think it can be organized like this, what do you think?

JSUYA: If SetFloatingMenuSupport() is not called first, when the right button is pressed, the FloatingMenu is displayed first while the ToastPopup is loading, and then SetFloatingMenuSupport() is executed. 

Of course, even with this approach, there were cases where the FloatingMenu appeared 1-2 times out of 10. However, since we couldn't control the global event of the right-click button on the FloatingMenu, I had no choice but to call SetFloatingMenuSupport() first.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -47,14 +53,107 @@ FlutterPointerDeviceKind ToFlutterDeviceKind(const Ecore_Device* dev) {
   }
 }
 
+#ifdef TV_PROFILE
+time_t GetBootTimeEpoch() {
+  struct timespec now, boottime;
```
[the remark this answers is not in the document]
JSUYA: done

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: I added the MouseSupport enum and clarified the function definitions.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -47,14 +53,107 @@ FlutterPointerDeviceKind ToFlutterDeviceKind(const Ecore_Device* dev) {
...
+    if (ret != APP_CONTROL_ERROR_NONE || !app_id) {
+      FT_LOG(Error) << "Fail to get app id.";
+      return false;
+    }
+
+    std::ostringstream boot_time_buffer;
+    boot_time_buffer << boot_time;
+    std::string preference_key =
+        std::string(kSysPointingDeviceSupportToastSharedPreferenceKey) + "/" +
+        app_id + "/" + boot_time_buffer.str();
```
xiaowei-guan, line 87: app_id must be released using free().

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
```diff
@@ -47,14 +53,107 @@ FlutterPointerDeviceKind ToFlutterDeviceKind(const Ecore_Device* dev) {
...
+
+  if (boot_time != -1) {
+    char* app_id = nullptr;
+    int ret = app_get_id(&app_id);
+    if (ret != APP_CONTROL_ERROR_NONE || !app_id) {
+      FT_LOG(Error) << "Fail to get app id.";
+      return false;
+    }
+
+    std::ostringstream boot_time_buffer;
```
xiaowei-guan, line 83: Line 73 ~ 87, can we extract a new method for preference_key?

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: Thank you for review. Removed duplication of Toast code.
If both are false, only SetFloatingMenuSupport() should be called.

flutter/shell/platform/tizen/tizen_window_ecore_wl2.cc
JSUYA: I added GetPreferenceKey() method.

PR #167, opened 2026-04-02
Add gdbus callback for Aurum (accessibility status monitoring)

seungsoo47:
- Connect to org.a11y.Bus to check Accessibility state changes
- Monitor IsEnabled and ScreenReaderEnabled property to properly enable/disable semantics

### DIFF
--- a/flutter/shell/platform/tizen/accessibility_settings.cc
+++ b/flutter/shell/platform/tizen/accessibility_settings.cc
@@ -15,23 +15,68 @@
 namespace flutter {
 
 AccessibilitySettings::AccessibilitySettings(FlutterTizenEngine* engine)
-    : engine_(engine) {
-  bool tts_enabled = false;
+    : engine_(engine),
+      tts_enabled_(false),
+      a11y_enabled_(false),
+      a11y_proxy_(nullptr),
+      cancellable_(g_cancellable_new()) {
+  InitializeScreenReaderMonitoring();
+  InitializeA11yMonitoring();
+  UpdateSemanticsState();
+  InitializeHighContrastMonitoring();
+}
+
+void AccessibilitySettings::InitializeScreenReaderMonitoring() {
   int ret = system_settings_get_value_bool(
-      SYSTEM_SETTINGS_KEY_ACCESSIBILITY_TTS, &tts_enabled);
+      SYSTEM_SETTINGS_KEY_ACCESSIBILITY_TTS, &tts_enabled_);
   if (ret != SYSTEM_SETTINGS_ERROR_NONE) {
     FT_LOG(Error) << "Failed to get value of accessibility tts.";
   }
-  if (tts_enabled) {
-    screen_reader_enabled_ = tts_enabled;
-    engine_->SetSemanticsEnabled(tts_enabled);
-  }
   system_settings_set_changed_cb(SYSTEM_SETTINGS_KEY_ACCESSIBILITY_TTS,
                                  OnScreenReaderStateChanged, this);
+}
 
+void AccessibilitySettings::InitializeA11yMonitoring() {
+  GError* error = nullptr;
+
+  a11y_proxy_ =
+      g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SESSION, G_DBUS_PROXY_FLAGS_NONE,
+                                    nullptr,            // GDBusInterfaceInfo
+                                    "org.a11y.Bus",     // name
+                                    "/org/a11y/bus",    // object path
+                                    "org.a11y.Status",  // interface
+                                    cancellable_,       // GCancellable
+                                    &error);            // GError
+
+  if (error) {
+    FT_LOG(Error) << "Failed to create a11y proxy: " << error->message;
+    g_error_free(error);
+    return;
+  }
+
+  GVariant* result = g_dbus_proxy_get_cached_property(a11y_proxy_, "IsEnabled");
+
+  if (error) {
+    FT_LOG(Error) << "Failed to get initial a11y state: " << error->message;
+    g_error_free(error);
+  } else {
+    a11y_enabled_ = g_variant_get_boolean(result);
+    g_variant_unref(result);
+    FT_LOG(Info) << "Initial a11y state: "
+                 << (a11y_enabled_ ? "enabled" : "disabled");
+  }
+
+  g_signal_connect_data(a11y_proxy_, "g-properties-changed",
+                        G_CALLBACK(OnA11yPropertiesChanged), this, nullptr,
+                        G_CONNECT_AFTER);
+
+  FT_LOG(Info) << "Connected to a11y status changes via GDBusProxy";
+}
+
+void AccessibilitySettings::InitializeHighContrastMonitoring() {
 #ifdef TV_PROFILE
   int high_contrast = 0;
-  ret = system_settings_get_value_int(
+  int ret = system_settings_get_value_int(
       SYSTEM_SETTINGS_KEY_ACCESSIBILITY_HIGHCONTRAST, &high_contrast);
   if (ret != SYSTEM_SETTINGS_ERROR_NONE) {
     FT_LOG(Error) << "Failed to get value of accessibility high contrast.";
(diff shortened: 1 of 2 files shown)


flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -15,23 +15,68 @@
...
+    return;
+  }
+
+  GVariant* result = g_dbus_proxy_get_cached_property(a11y_proxy_, "IsEnabled");
+
+  if (error) {
+    FT_LOG(Error) << "Failed to get initial a11y state: " << error->message;
+    g_error_free(error);
+  } else {
+    a11y_enabled_ = g_variant_get_boolean(result);
```
JSUYA: A null check for the result value is required as on line 170. Since g_dbus_proxy_get_cached_property() does not handle `error` variables, lines 57–63 need to be check again.

Please check gemini's review.
https://github.com/flutter-tizen/embedder/pull/167#discussion_r3025521737

flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -15,23 +15,68 @@
...
+  a11y_proxy_ =
+      g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SESSION, G_DBUS_PROXY_FLAGS_NONE,
+                                    nullptr,            // GDBusInterfaceInfo
+                                    "org.a11y.Bus",     // name
+                                    "/org/a11y/bus",    // object path
+                                    "org.a11y.Status",  // interface
+                                    cancellable_,       // GCancellable
+                                    &error);            // GError
+
+  if (error) {
```
JSUYA: Since the `error` argument of g_dbus_proxy_new_for_bus_sync() is assigned a value only when an error occurs, there may be no issue with the operation, but it is better to check the a11y_proxy value because it is to verify whether g_dbus_proxy_new_for_bus_sync() is work properly.
https://docs.gtk.org/gio/ctor.DBusProxy.new_for_bus_sync.html

flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -15,23 +15,68 @@
...
+    FT_LOG(Error) << "Failed to create a11y proxy: " << error->message;
+    g_error_free(error);
+    return;
+  }
+
+  GVariant* result = g_dbus_proxy_get_cached_property(a11y_proxy_, "IsEnabled");
+
+  if (error) {
```
JSUYA: like the comment above, you need to check the `result` value.

flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -15,23 +15,68 @@
...
+    g_variant_unref(result);
+    FT_LOG(Info) << "Initial a11y state: "
+                 << (a11y_enabled_ ? "enabled" : "disabled");
+  }
+
+  g_signal_connect_data(a11y_proxy_, "g-properties-changed",
+                        G_CALLBACK(OnA11yPropertiesChanged), this, nullptr,
+                        G_CONNECT_AFTER);
+
+  FT_LOG(Info) << "Connected to a11y status changes via GDBusProxy";
```
JSUYA: It seems like an unnecessary log.

flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -74,16 +144,39 @@ void AccessibilitySettings::OnScreenReaderStateChanged(
...
     FT_LOG(Error) << "Failed to get value of accessibility tts.";
     return;
   }
 
-  if (enabled != self->screen_reader_enabled_) {
-    self->screen_reader_enabled_ = enabled;
-    self->engine_->SetSemanticsEnabled(enabled);
+  if (old_tts_enabled != self->tts_enabled_) {
+    FT_LOG(Info) << "TTS state changed to: "
+                 << (self->tts_enabled_ ? "enabled" : "disabled");
```
JSUYA: I think there are too many unnecessary logs.
If logs are needed, the logs from UpdateSemanticsState() seem sufficient.

flutter/shell/platform/tizen/accessibility_settings.h
```diff
@@ -17,13 +18,27 @@ class AccessibilitySettings {
...
+  static void OnA11yPropertiesChanged(
+      GDBusProxy* proxy,
+      GVariant* changed_properties,
+      const gchar* const* invalidated_properties,
+      gpointer user_data);
 
   [[maybe_unused]] FlutterTizenEngine* engine_;
   [[maybe_unused]] bool screen_reader_enabled_ = false;
+  bool tts_enabled_ = false;
+  bool a11y_enabled_ = false;
```
JSUYA: I think these are redundant.
Generally, when the value of db/setting/accessibility/tts changes's vconf, the at-spi daemon executes at-spi-related services and also runs the "org.a11y.Bus" bus.

With the current configuration, it is expected that enabling accessibility features on a TV (or RPi device) will cause the callback to be invoked twice, resulting in a two-fold call to SetSemanticsEnabled().

Is there a specific reason why this must be done twice?
It would be helpful to refer to the implementation details of Dali or NUI.

flutter/shell/platform/tizen/accessibility_settings.cc
```diff
@@ -15,23 +15,68 @@
...
-  }
   system_settings_set_changed_cb(SYSTEM_SETTINGS_KEY_ACCESSIBILITY_TTS,
                                  OnScreenReaderStateChanged, this);
+}
+
+void AccessibilitySettings::InitializeA11yMonitoring() {
+  GError* error = nullptr;
+
+  a11y_proxy_ =
+      g_dbus_proxy_new_for_bus_sync(G_BUS_TYPE_SESSION, G_DBUS_PROXY_FLAGS_NONE,
```
JSUYA: g_dbus_proxy_new_for_bus_sync() is a synchronous call. Calling it in the constructor may affect embedder execution performance.

flutter/shell/platform/tizen/accessibility_settings.h
seungsoo47: As you know, we have been verifying TTS (which I guess that it is similar to dbus's ScreenReaderEnabled property used in Dali. Dali also implements dbus's isEnabled property). 
Recently, the QAD team requested Aurum support, and this PR includes the isEnabled property of dbus to implement the request.
For reference, you can detect dbus's isEnabled property when the Aurum App is turned on. However, in this implementation, the callback function as you mentioned is not called twice.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Oh. It's my mistake. I will update it.
Thank you.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Ok. I will update it.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Done

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Done.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Done.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Done.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: Done.

flutter/shell/platform/tizen/accessibility_settings.h
seungsoo47: I have updated it.

flutter/shell/platform/tizen/accessibility_settings.cc
seungsoo47: I have updated it.

JSUYA (approved):

PR #169, opened 2026-05-12
Use exact trig values for rotation transform in OnRotate

JSUYA:
cos/sin of pi multiples leaves ~1e-16 dust on entries that mathematically equal 0 or +-1. The dust scales y=1080 to 1079.99999... and shifts the bottom/right edge by 1 px in rasterizer paths that floor the transformed vertex. SendInitialGeometry calls OnRotate(degree=0) at startup, so non-rotating Skia + GLES apps were also affected and the player overlay (avplay videohole) appeared 1-2 px below adjacent Flutter widgets at the screen's bottom edge. Impeller's GLES backend short-circuits near-identity transforms and did not exhibit the shift.

Use exact 0 / +-1 / width / height entries for the cardinal angles (0, 90, 180, 270), and keep the cos/sin computation in the default branch so arbitrary rotation degrees remain supported.

https://github.com/flutter-tizen/flutter-tizen/issues/764

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen_view.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_view.cc
@@ -125,24 +125,30 @@ void FlutterTizenView::OnRotate(int32_t degree) {
   int32_t width = geometry.width;
   int32_t height = geometry.height;
   if (dynamic_cast<TizenRendererEgl*>(engine_->renderer())) {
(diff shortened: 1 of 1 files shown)


xiaowei-guan (approved):

xiaowei-guan (approved):
Verified on TV device

PR #170, opened 2026-05-15
Support `tizen-core` window

JSUYA:
Adds an optional window / input-method backend on tizen-core-wl + tizen-core-imf instead of Ecore / ecore_wl2, so the embedder keeps working on images where EFL is deprecated. Off by default; --use-tcore (GN use_tcore), rejected below Tizen 10.0.

`tools/gn ... --api-version 11.0 --use-tcore --target-dir tizen_arm64_11.0`

Artifact list is unchanged: the same six libraries are rebuilt from the same template("embedder") with different backend sources and libs. No new target, no new public API.

- libflutter_tizen_common.so
- libflutter_tizen_mobile.so
- libflutter_tizen_tv.so
- libflutter_tizen_{common,mobile,tv}_experimental.so — same as above + the Vulkan renderer (FLUTTER_TIZEN_EXPERIMENTAL, volk)

*Only exercised on Tizen 11.


JSUYA:
#195

PR #171, opened 2026-05-26
Fix event loop re-entrancy issue on merged thread policy

seungsoo47:
This PR resolves an event loop re-entrancy issue under the merged UI/Platform thread policy by using a local variable to prevent memory corruption and task dispatch failures.

### DIFF
--- a/flutter/shell/platform/tizen/tizen_event_loop.cc
+++ b/flutter/shell/platform/tizen/tizen_event_loop.cc
@@ -104,7 +104,6 @@ void TizenPlatformEventLoop::OnTaskExpired() {
   {
     std::lock_guard<std::mutex> lock(expired_tasks_mutex_);
     local_expired_tasks = std::move(expired_tasks_);
-    expired_tasks_.clear();
   }
 
   for (const Task& task : local_expired_tasks) {
--- a/flutter/shell/platform/tizen/tizen_event_loop.cc
+++ b/flutter/shell/platform/tizen/tizen_event_loop.cc
@@ -100,10 +100,15 @@ TizenPlatformEventLoop::TizenPlatformEventLoop(
 TizenPlatformEventLoop::~TizenPlatformEventLoop() {}
 
 void TizenPlatformEventLoop::OnTaskExpired() {
-  for (const Task& task : expired_tasks_) {
+  std::vector<Task> local_expired_tasks;
+  {
+    std::lock_guard<std::mutex> lock(expired_tasks_mutex_);
+    local_expired_tasks = std::move(expired_tasks_);
+  }
+
+  for (const Task& task : local_expired_tasks) {
     on_task_expired_(&task.task);
   }
-  expired_tasks_.clear();
 }
 
 }  // namespace flutter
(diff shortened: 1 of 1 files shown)


flutter/shell/platform/tizen/tizen_event_loop.cc
```diff
@@ -100,10 +100,16 @@ TizenPlatformEventLoop::TizenPlatformEventLoop(
 TizenPlatformEventLoop::~TizenPlatformEventLoop() {}
 
 void TizenPlatformEventLoop::OnTaskExpired() {
-  for (const Task& task : expired_tasks_) {
+  std::vector<Task> local_expired_tasks;
+  {
+    std::lock_guard<std::mutex> lock(expired_tasks_mutex_);
+    local_expired_tasks = std::move(expired_tasks_);
+    expired_tasks_.clear();
```
JSUYA: It looks like you don't need to call clear() on `expired_tasks_` because it has already been moved.

flutter/shell/platform/tizen/tizen_event_loop.cc
```diff
@@ -100,10 +100,16 @@ TizenPlatformEventLoop::TizenPlatformEventLoop(
 TizenPlatformEventLoop::~TizenPlatformEventLoop() {}
 
 void TizenPlatformEventLoop::OnTaskExpired() {
-  for (const Task& task : expired_tasks_) {
+  std::vector<Task> local_expired_tasks;
```
JSUYA, line 103: How about moving it inside the mutex? It is used only there.

flutter/shell/platform/tizen/tizen_event_loop.cc
seungsoo47: It seems the reason Gemini suggested is due to the "valid but unspecified state" issue.
And I checked and found that since C++17, the state after the move is guaranteed to be empty.
Therefore I will remove the relevant code because we are using C++17.

flutter/shell/platform/tizen/tizen_event_loop.cc
seungsoo47: What does it mean? Can you explain it in more detail?

JSUYA: What I mean is this.
```
  {
    std::lock_guard<std::mutex> lock(expired_tasks_mutex_);
    std::vector<Task> local_expired_tasks;
    local_expired_tasks = std::move(expired_tasks_);
    expired_tasks_.clear();
  }
```
It doesn't seem important. just ignore it.

seungsoo47: The local_expired_tasks is used in the code below:
```
  for (const Task& task : local_expired_tasks) {
    on_task_expired_(&task.task);
  }
```

JSUYA: oh i'm sorry. my mistake.

JSUYA (approved):

PR #182, opened 2026-07-07
feat(texture): Tear down external textures on the render thread

seungsoo47:
UnregisterTexture() previously destroyed the ExternalTexture (running glDeleteTextures) synchronously on the calling platform thread. That could race with an in-flight PopulateGLTexture() on the render thread and free the backing TBM buffer while the GPU still referenced it, causing heap corruption and "tbm_bo_free ... lock_cnt" crashes on the SW backend.

- Add a completion-callback parameter to UnregisterTexture, plumbed through FlutterDesktopTextureRegistrarUnregisterExternalTexture, so callers learn when teardown has actually completed.
- Add FlutterTizenEngine::PostRenderThreadTask(), a thin wrapper over the embedder API's PostRenderThreadTask.
- UnregisterTexture now removes the texture from the map (blocking new lookups), then posts its destruction to the render thread. The engine runs that task only after any in-flight frame callback has completed, so it no longer races; glDeleteTextures runs on the thread that owns the GL context, and the completion callback fires only once teardown is done.

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen_engine.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_engine.cc
@@ -412,6 +412,24 @@ bool FlutterTizenEngine::MarkExternalTextureFrameAvailable(int64_t texture_id) {
               engine_, texture_id) == kSuccess);
 }
 
+void FlutterTizenEngine::PostRenderThreadTask(std::function<void()> task) {
+  if (!engine_) {
+    // The engine is shutting down (or never started); posting is unsafe.
+    // Run inline so the task's cleanup still happens rather than leaking.
+    task();
+    return;
+  }
+  auto* heap_task = new std::function<void()>(std::move(task));
+  embedder_api_.PostRenderThreadTask(
+      engine_,
+      [](void* data) {
+        auto* fn = static_cast<std::function<void()>*>(data);
+        (*fn)();
+        delete fn;
+      },
+      heap_task);
+}
+
 void FlutterTizenEngine::UpdateAccessibilityFeatures(bool invert_colors,
                                                      bool high_contrast) {
   int32_t flags = 0;
--- a/flutter/shell/platform/tizen/flutter_tizen_engine.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_engine.cc
@@ -420,14 +420,19 @@ void FlutterTizenEngine::PostRenderThreadTask(std::function<void()> task) {
     return;
   }
   auto* heap_task = new std::function<void()>(std::move(task));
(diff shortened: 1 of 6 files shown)


flutter/shell/platform/tizen/flutter_tizen_engine.cc
```diff
@@ -412,6 +412,24 @@ bool FlutterTizenEngine::MarkExternalTextureFrameAvailable(int64_t texture_id) {
...
 
+void FlutterTizenEngine::PostRenderThreadTask(std::function<void()> task) {
+  if (!engine_) {
+    // The engine is shutting down (or never started); posting is unsafe.
+    // Run inline so the task's cleanup still happens rather than leaking.
+    task();
+    return;
+  }
+  auto* heap_task = new std::function<void()>(std::move(task));
+  embedder_api_.PostRenderThreadTask(
```
xiaowei-guan: If `embedder_api_.PostRenderThreadTask()` returns an error (not `kSuccess`), the `heap_task` is never scheduled and never deleted, causing a memory leak. The return value is not checked.

flutter/shell/platform/tizen/flutter_tizen_engine.h
```diff
@@ -171,6 +172,13 @@ class FlutterTizenEngine {
   // given |texture_id|.
   bool MarkExternalTextureFrameAvailable(int64_t texture_id);
 
+  // Posts |task| onto the Flutter render (raster) thread. May be called from
+  // any thread while the engine is running. Used to tear down external-texture
+  // GPU resources on the thread that owns the GL context, after any in-flight
+  // frame callback for that texture has completed. Does nothing if the engine
+  // is not running.
+  void PostRenderThreadTask(std::function<void()> task);
```
xiaowei-guan, line 181: Suggest change PostRenderThreadTask return value from void to bool, If calling embedder_api_.PostRenderThreadTask() fails, the caller is responsible for handling the failure.

xiaowei-guan (dismissed):

flutter/shell/platform/tizen/flutter_tizen_engine.cc
seungsoo47: PostRenderThreadTask() has been modified to check its return value and execute the task inline if posting fails.

flutter/shell/platform/tizen/flutter_tizen_engine.h
seungsoo47: Kept void. On failure PostRenderThreadTask now runs the task inline internally (fixed in 46c8f93).

xiaowei-guan (approved):

PR #185, opened 2026-07-15
Support vulkan pipeline cache

### LINKED ISSUE #181: Support Vulkan pipeline cache
To improve startup performance in later runs of the app, Impeller can use Vulkan's pipeline cache to cache some shader pipeline data to disk. To enable this feature, the embedder must configure Impeller Vulkan's pipeline cache directory ([ContextVK::Settings::cache_directory](https://github.com/flutter/flutter/blob/496e76f95b64889b03234f7d450aea0fdddf0c8e/engine/src/flutter/impeller/renderer/backend/vulkan/context_vk.h#L82)).
https://github.com/flutter/flutter/issues/186374

xiaowei-guan:
Fix https://github.com/flutter-tizen/embedder/issues/181

The cache is now successfully generated and loaded when the app starts.

Currently, no significant speedup effect from VkPipelineCache is observed on RPI4, largely due to the small number of pipelines, simple structure, or small initialization scale.

Due to issues with the Vulkan driver on TVs, this PR cannot be verified on TVs.

### DIFF
--- a/flutter/shell/platform/tizen/tizen_renderer_vulkan.cc
+++ b/flutter/shell/platform/tizen/tizen_renderer_vulkan.cc
@@ -4,6 +4,7 @@
 
 #include "flutter/shell/platform/tizen/tizen_renderer_vulkan.h"
 
+#include <app.h>
 #include <stddef.h>
 #include <optional>
 #include "flutter/shell/platform/tizen/external_texture_pixel_vulkan.h"
@@ -49,6 +50,11 @@ TizenRendererVulkan::TizenRendererVulkan(TizenViewBase* view) {
   FT_LOG(Info) << "Vulkan version: " << VK_VERSION_MAJOR(version) << "."
                << VK_VERSION_MINOR(version) << "." << VK_VERSION_PATCH(version);
 
+  char* cache_path = app_get_cache_path();
+  if (cache_path) {
+    cache_path_ = cache_path;
+    free(cache_path);
+  }
   InitVulkan(view);
 }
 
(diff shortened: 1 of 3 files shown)


flutter/shell/platform/tizen/tizen_renderer_vulkan.cc
```diff
@@ -200,6 +206,7 @@ FlutterRendererConfig TizenRendererVulkan::GetRendererConfig() {
     return engine->texture_registrar()->PopulateVulkanTexture(texture_id, width,
                                                               height, texture);
   };
+  config.vulkan.cache_path = cache_path_.c_str();
```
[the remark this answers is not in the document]
xiaowei-guan: Done

flutter/shell/platform/tizen/tizen_renderer_vulkan.cc
```diff
@@ -49,6 +50,11 @@ TizenRendererVulkan::TizenRendererVulkan(TizenViewBase* view) {
   FT_LOG(Info) << "Vulkan version: " << VK_VERSION_MAJOR(version) << "."
                << VK_VERSION_MINOR(version) << "." << VK_VERSION_PATCH(version);
 
+  char* cache_path = app_get_cache_path();
+  if (cache_path) {
+    cache_path_ = cache_path;
+    free(cache_path);
+  }
```
JSUYA, line 60:
```
} else {
  FT_LOG(Warn)
      << "app_get_cache_path() failed. Vulkan pipeline cache is disabled.";
}
```
Hmm... I'm not sure if there are actually cases where app_get_cache_path() fails, but since the cache won't work if the path doesn't exist, I think it's a good idea to log it.

JSUYA (approved):

flutter/shell/platform/tizen/tizen_renderer_vulkan.cc
xiaowei-guan: Done

PR #188, opened 2026-07-24
Send view focus events and handle focus change requests

### LINKED ISSUE #187: Support view focus events
The engine's view-focus embedder API (`FlutterEngineSendViewFocusEvent`, `view_focus_change_request_callback`) is already in our `embedder.h`, but the Tizen embedder doesn't wire it up. 

So `PlatformDispatcher.onViewFocusChange` and `requestViewFocusChange` don't work on Tizen.

related flutter PR : flutter/flutter#164296

JSUYA:
The engine's view-focus embedder API (FlutterEngineSendViewFocusEvent, view_focus_change_request_callback) is declared in embedder.h but was never wired up, so PlatformDispatcher.onViewFocusChange and requestViewFocusChange did not work on Tizen.

Since the embedder has no native window focus event handler, the app lifecycle is used as a proxy for view focus: kFocused is sent to the engine when the app is resumed, and kUnfocused when it becomes inactive or paused. Focus change requests from the framework are handled by activating the window. Windows created with
focusable = false are marked with ecore_wl2_window_focus_skip_set and can never receive native key focus, so kFocused events and window activation are skipped for them.

Fixes https://github.com/flutter-tizen/embedder/issues/187

### DIFF
--- a/flutter/shell/platform/tizen/BUILD.gn
+++ b/flutter/shell/platform/tizen/BUILD.gn
@@ -235,6 +235,7 @@ executable("flutter_tizen_unittests") {
     "flutter_project_bundle_unittests.cc",
     "flutter_tizen_engine_unittest.cc",
     "flutter_tizen_texture_registrar_unittests.cc",
+    "flutter_tizen_view_unittests.cc",
   ]
 
   ldflags = [ "-Wl,--unresolved-symbols=ignore-in-shared-libs" ]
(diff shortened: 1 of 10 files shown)


seungsoo47:
This patch looks good, but what do you think about adding unit tests?

JSUYA:
> This patch looks good, but what do you think about adding unit tests?

I added test code. thanks

seungsoo47 (approved):

PR #191, opened 2026-08-04
Support gcc 10+ toolchains in accessibility headers

JSUYA:
gcc 9.2.0 (Tizen 6.0/6.5/8.0) compiles these headers as-is, gcc 10 and later do not. Verified clean at -std=c++17 -Werror against gcc 9.2.0 (arm, arm64, x64 sysroots), libstdc++ 10, 11, 12, 13 and gcc 14.2.0 (Tizen 11.0 arm64).

AXEventGenerator::Iterator now matches the current Chromium implementation, which spells out the five iterator_traits typedefs instead of deriving from the deprecated std::iterator: https://source.chromium.org/chromium/chromium/src/+/main:ui/accessibility/ax_event_generator.h

### DIFF
--- a/flutter/third_party/accessibility/ax/ax_active_popup.h
+++ b/flutter/third_party/accessibility/ax/ax_active_popup.h
@@ -5,6 +5,7 @@
 #ifndef UI_ACCESSIBILITY_AX_ACTIVE_POPUP_H_
 #define UI_ACCESSIBILITY_AX_ACTIVE_POPUP_H_
 
+#include <cstdint>
 #include <optional>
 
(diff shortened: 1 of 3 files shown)


seungsoo47:
It seems that AXRange::Iterator in ax_range.h uses std::iterator.
Please revise the necessary parts after reviewing.

JSUYA:
> It seems that AXRange::Iterator in ax_range.h uses std::iterator. Please revise the necessary parts after reviewing.

I added more commit. thanks

seungsoo47 (approved):

Issue #146, opened 2026-02-04
Implement `Tizen Core` embedder


JSUYA:
Currently, our embedder is built on EFL's ecore, ecore_imf, ecore_input, ecore_wl2, eina, and eldbus libraries.
(Dependencies on evas and elm have already been removed.)

The EFL library was deprecated in Tizen 10.0, and according to Tizen policy, its API will be removed +2 version later.
(https://review.tizen.org/gerrit/c/platform/upstream/efl/+/321724)
(https://review.tizen.org/gerrit/c/platform/upstream/efl/+/321255)
(https://review.tizen.org/gerrit/c/platform/upstream/efl/+/324645)

Tizen will transition from EFL's ecore to TizenCore (tcore).
https://docs.tizen.org/application/native/guides/applications/tizen-core/tizen-core/

Therefore, a separate embedder that implements code currently written in ecore and ecore_imf in tcore must be prepared.
(This change will apply to the common profile first, and the TV profile is not yet planned.)

Action Items
- eldbus will be replaced to gdbus.
- eina will be replaced to std::vector (maybe?)
- ecore api will replaced to tcore
- ecore_wl2 api will replaced to tcore_wayland
- ecore used in plugins will be replaced with tcore.
- Below target plugins will add tcore-based C++ implementations:
flutter-tizen/plugins/audioplayers
flutter-tizen/plugins/geolocator
flutter-tizen/plugins/image_picker
flutter-tizen/plugins/permission_handler
flutter-tizen/plugins/tizen_app_manager
flutter-tizen/plugins/tizen_package_manager
flutter-tizen/plugins/tizen_window_manager
flutter-tizen/plugins/video_player
flutter-tizen/plugins/video_player_avplay
flutter-tizen/plugins/video_player_videohole
flutter-tizen/plugins/wakelock_plus
flutter-tizen/plugins/webview_flutter_lwe

JSUYA:
I've uploaded PoC commit for tizen-core-wl to our internal git (f-project/embedder).
I'd like to share some additional work items identified during this process.

- `devel-api` of the dali package, used to implement the NUIFlutterView, has been separated into the `integration-devel` rpm in Tizen 11. Therefore, sysroots for Tizen 11 and later may require separate configuration.
- `TizenEventLoop`, implemented using `Ecore_Pipe`, may need to be implemented using std::thread. (like this https://github.com/flutter-tizen/embedder/pull/155)
[...]

PR #192, opened 2026-08-07
Replace ecore_pipe to glib api in TizenEventLoop

JSUYA:
Replace Ecore pipes and per-task timers with a single GLib source that tracks the earliest Flutter task on the Tizen main context.

Preserve delayed execution, posting order, worker-thread scheduling, re-entrant callbacks, and source cleanup. Remove the remaining Ecore dependency from display refresh reporting by using the existing 60 Hz constant.

https://github.com/flutter-tizen/embedder/issues/146

### DIFF
--- a/flutter/shell/platform/tizen/flutter_tizen_display_monitor.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_display_monitor.cc
@@ -3,7 +3,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include <Ecore.h>
 #include <system_info.h>
 
 #include "flutter/shell/platform/tizen/flutter_tizen_display_monitor.h"
@@ -29,12 +28,7 @@ void FlutterTizenDisplayMonitor::UpdateDisplays() {
   display.display_id = 0;
   display.single_display = true;
 
-  double fps = ecore_animator_frametime_get();
-  if (fps <= 0.0) {
-    display.refresh_rate = 0.0;
-  } else {
-    display.refresh_rate = 1 / fps;
-  }
+  display.refresh_rate = 60.0;
 
   int32_t width = 0, height = 0, dpi = 0;
   FlutterTizenView* view = engine_->view();
--- a/flutter/shell/platform/tizen/flutter_tizen_display_monitor.cc
+++ b/flutter/shell/platform/tizen/flutter_tizen_display_monitor.cc
@@ -28,6 +28,8 @@ void FlutterTizenDisplayMonitor::UpdateDisplays() {
   display.display_id = 0;
   display.single_display = true;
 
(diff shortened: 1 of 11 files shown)


JSUYA:
@codex review

flutter/shell/platform/tizen/flutter_tizen_display_monitor.cc
```diff
@@ -29,12 +28,7 @@ void FlutterTizenDisplayMonitor::UpdateDisplays() {
   display.display_id = 0;
   display.single_display = true;
 
-  double fps = ecore_animator_frametime_get();
-  if (fps <= 0.0) {
-    display.refresh_rate = 0.0;
-  } else {
-    display.refresh_rate = 1 / fps;
-  }
+  display.refresh_rate = 60.0;
```
seungsoo47, line 33: The Linux embedder (fl_display_monitor.cc) gets the real per-monitor refresh_rate via gdk_monitor_get_refresh_rate() and updates it on hotplug. GDK isn't available on Tizen, but the same value is obtainable by directly parsing the refresh field from the wl_output.mode event - no Ecore needed. Could you at least file a follow-up issue for the hardcoded 60.0 instead of leaving it silent?

JSUYA: `ecore_animator_frametime_get()` returns the default value of 1 / 60.0f unless otherwise set. Therefore, apply 60.0 to match existing behavior.

seungsoo47: You're right that ecore_animator_frametime_get() defaults to 60, but it doesn't actually reflect the real monitor's frame rate, so this will need to be fixed eventually. Even if not right now, could we at least leave a follow-up issue for it?

JSUYA: um... https://github.com/flutter-tizen/embedder/issues/193 I have made an issue. However, I am not sure if we will be able to resolve it.

JSUYA:
Performance: master (ecore_pipe) vs this PR (GSource)

  Measured on the x64 emulator (debug, 2 runs each) with no-op platform channel calls:

  | Metric | master | This PR | Delta |
  |---|---|---|---|
  | Channel round-trip p50 | 98–104 µs | 80–87 µs | ~15–20% faster |
  | 5000-call burst wall / CPU | 297 ms / ~360 ms | ~205 ms / ~255 ms | ~30% faster, ~30% less CPU |
  | Timer lateness, fps | — | — | unchanged |

  The old loop issued one `ecore_pipe_write()` syscall per immediate `PostTask()`;
  `g_source_set_ready_time()` coalesces bursts into a single wakeup. Emulator numbers,
  but the gain comes from removing a per-post syscall, so it should hold on devices.

xiaowei-guan (approved):

PR #195, opened 2026-08-21
Support tizen-core window

JSUYA:
Adds `--use-tcore`, which swaps the window, input-method, clipboard and softkey backends from EFL (`ecore_wl2` / `ecore_imf`) to `tizen-core-wl` / `tizen-core-imf`, so the embedder keeps working on images where EFL is deprecated. Off by default.

```
$) tools/generate_sysroot.py --api-version 11.0 --arch arm64   # adds the tizen-core packages
$) tools/gn --target-arch arm64 --api-version 11.0 --use-tcore --target-dir tizen_arm64_11.0
```

Requires api-version 11.0 or higher — tizen-core-wl / tizen-core-imf ship only on Tizen 11.0 and every API used here is `@since_tizen 11.0`.

- The TV-only `libvd-win-util.so` helpers moved into a new `TizenWindowUtil`, so `tizen_window_ecore_wl2.cc` loses its own `dlopen`/`dlsym` block (-102/+18). Intended to be behaviour-neutral.
- `FlutterDesktopViewGetNativeHandle()` returns a different type in a tcore build (`tizen_core_wl_window_h` vs `Ecore_Wl2_Window*`). Signature and exported symbols are unchanged, but plugins that cast the result must handle it.

### DIFF
--- a/flutter/shell/platform/tizen/tizen_input_method_context_tcore.cc
+++ b/flutter/shell/platform/tizen/tizen_input_method_context_tcore.cc
@@ -1,0 +1,453 @@
+// Copyright 2026 Samsung Electronics Co., Ltd. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "tizen_input_method_context_tcore.h"
+
+#include <tizen_core_wl.h>
+
+#include "flutter/shell/platform/tizen/logger.h"
+
+namespace {
+
+tizen_core_imf_input_panel_layout_e TextInputTypeToImfInputPanelLayout(
+    const std::string& text_input_type) {
+  if (text_input_type == "TextInputType.text" ||
+      text_input_type == "TextInputType.multiline") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_NORMAL;
+  } else if (text_input_type == "TextInputType.number") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_NUMBERONLY;
+  } else if (text_input_type == "TextInputType.phone") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_PHONENUMBER;
+  } else if (text_input_type == "TextInputType.datetime") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_DATETIME;
+  } else if (text_input_type == "TextInputType.emailAddress" ||
+             text_input_type == "TextInputType.twitter") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_EMAIL;
+  } else if (text_input_type == "TextInputType.url" ||
+             text_input_type == "TextInputType.webSearch") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_URL;
+  } else if (text_input_type == "TextInputType.visiblePassword") {
+    return TIZEN_CORE_IMF_INPUT_PANEL_LAYOUT_PASSWORD;
+  } else {
+    FT_LOG(Warn) << "The requested input type " << text_input_type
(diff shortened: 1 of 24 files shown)


JSUYA:
https://github.com/flutter-tizen/embedder/pull/170

JSUYA:
@codex review

JSUYA (commented):
Three functional issues below. Checked against tizen-core-wl 0.1.43 and tizen-core-imf 0.2.17 source; not device-tested.

flutter/shell/platform/tizen/tizen_input_method_context_tcore.cc
```diff
@@ -0,0 +1,526 @@
...
+      event_callbacks_[TIZEN_CORE_IMF_CALLBACK_PREEDIT_END], this);
+
+  // pre-edit changed callback
+  event_callbacks_[TIZEN_CORE_IMF_CALLBACK_PREEDIT_CHANGED] =
+      [](tizen_core_imf_context_h ctx, void* event_info, void* data) {
+        auto* self = static_cast<TizenInputMethodContext*>(data);
+        if (self->on_preedit_changed_) {
+          char* str = nullptr;
+          int cursor_pos = 0;
+          tizen_core_imf_context_get_preedit_string(ctx, &str, nullptr, nullptr,
```
JSUYA: **[P1] Pass the required preedit outputs**

`get_preedit_string()` rejects null `attrs` or `attrs_count`, so this call always fails and composing text never reaches Flutter. Pass both output variables, check the result, and release attributes with `tizen_core_imf_preedit_attrs_destroy()`.

flutter/shell/platform/tizen/tizen_window_tcore_wl.cc
```diff
@@ -0,0 +1,871 @@
...
+                static_cast<tizen_core_wl_window_angle_e>(degree));
+            tizen_core_wl_window_send_rotation_change_done(
+                self->tcore_wl_window_, degree);
+          }
+        }
+      });
+
+  if (!is_vulkan_) {
+    AddEventListener(
+        TIZEN_CORE_WL_EVENT_WINDOW_CONFIGURE_COMPLETE,
```
JSUYA, line 397: **[P1] Handle subsequent window configurations**

`WINDOW_CONFIGURE_COMPLETE` signals initial surface configuration; later resizes emit `WINDOW_CONFIGURE`, leaving the EGL buffer and Flutter metrics unchanged. Listen for `WINDOW_CONFIGURE` and apply its geometry to the native window, EGL buffer, and `OnResize()`.

flutter/shell/platform/tizen/tizen_window_tcore_wl.cc
```diff
@@ -0,0 +1,871 @@
...
+            tizen_core_wl_event_mouse_button_get_position(ev, &x, &y);
+            unsigned int buttons = 0;
+            tizen_core_wl_event_mouse_button_get_buttons(ev, &buttons);
+            uint32_t timestamp = 0;
+            tizen_core_wl_event_input_base_get_timestamp(ev, &timestamp);
+            unsigned int touch_id = 0;
+            tizen_core_wl_event_mouse_button_get_touch_id(ev, &touch_id);
+            self->view_delegate_->OnPointerDown(
+                x, y, ToFlutterPointerButton(buttons), timestamp,
+                touch_id == 0 ? kFlutterPointerDeviceKindMouse
```
JSUYA, line 458: **[P2] Preserve touch device type for contact 0**

Tizen emits touch ID `0`, so this branch reports real touches as mouse input and prevents default drag scrolling. Carry reliable touch/mouse metadata through tcore and use it for down/up/move; 0.1.43 also assigns pointer metadata to ID `0`, so an embedder-only device lookup is insufficient.

seungsoo47 (approved):

PR #198, opened 2026-09-08
Add flutter tizen window api

xiaowei-guan:
We have provided a [window channel ](https://github.com/flutter-tizen/embedder/blob/main/flutter/shell/platform/tizen/channels/window_channel.h)for Dart, but the native side cannot call this interface. If call it from the Dart side would actually impact performance.

Related issue : https://github.com/flutter-tizen/plugins/issues/1138


JSUYA (commented):
Of course, calling the API directly rather than using a Method channel may be the fastest in performance.

However, since the Embedder looks as the owner of Window APIs, it is necessary to detail review whether the Embedder is truly responsible for operations related to the Window.

Additionally, the Embedder provides similar functionalities as Method channels, and the names of each channel share the common meaning of the methods.
I believe the performance is at an acceptable level, as Window APIs are not called frequently enough to significantly impact performance.

If additional Geometry-related APIs are needed, how about adding a Method channel and calling it using a window manager plugin?

xiaowei-guan:
> Of course, calling the API directly rather than using a Method channel may be the fastest in performance.
> 
> However, since the Embedder looks as the owner of Window APIs, it is necessary to detail review whether the Embedder is truly responsible for operations related to the Window.
> 
> Additionally, the Embedder provides similar functionalities as Method channels, and the names of each channel share the common meaning of the methods. I believe the performance is at an acceptable level, as Window APIs are not called frequently enough to significantly impact performance.
> 
> If additional Geometry-related APIs are needed, how about adding a Method channel and calling it using a window manager plugin?

If use Method channel to get window Geometry, we need to call GetGeometry  at dart side， and send the data to native side.

Issue #14, opened 2022-07-20
Key events when keep pressing an arrow key


WonyoungChoi:
When keep pressing an arrow key (up, down, left, right) of keyboard, the key event seems to occur only once. 

***Repro***
1. In any TextField, type any characters
2. Keep pressing an arrow key to move cursor continuously. 
3. But, the cursor moves single step. 

***Expected***
The cursor should be moved continuously like typing any other characters.

This isn't limited to TextField, it happens in general, like moving a Widget's focus. 

***Environment***
- Device : RPI4 
- Platform: tizen-unified_20220713.2_tizen-headed-armv7l
- Embedder: Latest

PR #199, opened 2026-09-08
Add arrow key repeat fallback for keyboards

JSUYA:
Some Tizen devices do not repeat arrow keys while they are held down. Repeat them here instead, using the repeat delay and rate the seat reports. If the device does send a repeat of its own, stop and let it through.

This follows the keyboard repeat in DALi's Windows backend. It uses a one-shot timer that re-arms on every tick, and reads a flag on the event to tell the device's own repeats from a fresh press.


JSUYA:
#14

seungsoo47:
The code looks good. Just wondering, have you tested it on Tizen 11.0?

JSUYA:
> The code looks good. Just wondering, have you tested it on Tizen 11.0?

Yes, I tested it on Tizen 11.0 RPI and the 10.1 emulator. Is there a problem?

seungsoo47:
> > The code looks good. Just wondering, have you tested it on Tizen 11.0?
> 
> Yes, I tested it on Tizen 11.0 RPI and the 10.1 emulator. Is there a problem?

I have just been curious.

seungsoo47 (approved):

